PR #20, opened 2019-06-20
Holoport preflight

pjkundert:
Attempt to run a HoloPortOS "Preflight" checklist, to check and repair any issues that have been detected since the last deployment of the `scripts/holoport-preflight.sh` script.

- The testing procedure doesn't seem to provide access to any new binaries provisioned in the nix pkgs stanza of base.nix; eg. `rsync`, `logger`.
- sshd.service and zerotier.service should probably be restarted, so new host keys and ZeroTier identity gets immediately created and used; this doesn't occur on automatic update (because these services are already running).

Also, changes the `systemd-activation` to `holochain-activation`, which gets its functionality from `scripts/holochain-activation.sh`.  Furthermore -- this is changed to a simple "oneshot" systemd service that is invoked by a `requires = ...` dependency from `holochain.service`.

Ensure that any service that may have its configuration modified by any phase of `holoport-preflight` is mentioned in the `before = ...` stanza, to ensure the Preflight configuration changes occur before that service starts!

Since this, of course, doesn't affect already-running services -- when we dynamically upgrade a host to a new version of holoportos, the affected services will continue to run their old configurations until restarted, or the next reboot.  If a service must be restarted, we have to do it manually.

This affects, specifically, the ZeroTier networking.  By wiping out the (default) identity.secret, we're going to lose access to the ZeroTier network, 'til the new identity is approved in ZeroTier, and entered into the cloudflare KV stores w/ the new IP address.

### DIFF
--- a/modules/base.nix
+++ b/modules/base.nix
@@ -176,6 +176,7 @@ in
         yarn
         zeromq4
         rsync
+        utillinux
       ];
       programs.bash.shellAliases = {
         htst = "${hptest}/bin/hptest";
--- a/modules/base.nix
+++ b/modules/base.nix
@@ -246,13 +246,15 @@ in
 
       # HoloPort Preflight checks must occur before ZeroTier and Holo / Holochain startup (or any
       # other service which may have its configuration modified by the holoport-preflight).
-      # Holochain Activation depends on running the "preflight" configuration checklist (to bring
-      # the basic HoloPort OS configuration up to standard), and then the Activation required for
-      # Holo / Holochain.  If either of these fail, the HoloPort will *not* be considered
-      # "Activated".
+      # Furthermore, these services must be restarted whenever the Preflight service is run (eg. on
+      # automatic HoloPortOS updates).  Holochain Activation depends on running the "preflight"
+      # configuration checklist (to bring the basic HoloPort OS configuration up to standard), and
+      # then the Activation required for Holo / Holochain.  If either of these fail, the HoloPort
+      # will *not* be considered "Activated".
       systemd.services.holoport-preflight = {
         enable          = true;
         before          = [ "zerotierone.service" "sshd.service" ];
+        requiredBy      = [ "zerotierone.service" "sshd.service" ];
         path            = [ pkgs.rsync pkgs.utillinux ];
         serviceConfig   = {
           Type          = "oneshot";
--- a/modules/base.nix
+++ b/modules/base.nix
@@ -176,7 +176,6 @@ in
         yarn
         zeromq4
         rsync
-        util-linux
       ];
       programs.bash.shellAliases = {
         htst = "${hptest}/bin/hptest";
--- a/modules/base.nix
+++ b/modules/base.nix
@@ -75,39 +75,14 @@ let
     echo "All tests have completed"
     cat hpplustest.txt | less
     '';
-  systemd-activation = pkgs.writeShellScriptBin "systemd-activation" ''
-    if [ ! -d /var/lib/holochain ] ;
-    then mkdir /var/lib/holochain; chown -R holochain:holochain /var/lib/holochain;
-    fi
 
-    if [ ! -d /home/holochain/.n3h ] ;
-    then mkdir /home/holochain/.n3h; chown -R holochain:holochain /home/holochain/.n3h;
-    fi
+  holoport-preflight = pkgs.writeShellScriptBin "holoport-preflight" (
+    builtins.readFile ../scripts/holoport-preflight.sh
+  );
+  holochain-activation = pkgs.writeShellScriptBin "holochain-activation" (
+    builtins.readFile ../scripts/holochain-activation.sh
+  );
 
-    if [ ! -d /home/holochain/.config/holochain/keys ] ;
-    then mkdir -p /home/holochain/.config/holochain/keys; chown -R holochain:holochain /home/holochain/.config/holochain/keys;
-    fi
-
-
-    if [ ! -f /var/lib/holochain/conductor-config.toml ];
-    then cat <<- EOF > /var/lib/holochain/conductor-config.toml
-    persistence_dir = "/var/lib/holochain"
-    agents = []
-    dnas = []
-    instances = []
-    interfaces = []
-    bridges = []
-
-    [logger]
-    type = "debug"
-
-    EOF
-    fi
-    chown holochain:holochain /var/lib/holochain/conductor-config.toml;
-    chown -R holochain:holochain /var/lib/holochain;
-    chmod 0775 /var/lib/holochain/conductor-config.toml;
-
-  '';
 in
 {
   options = {
@@ -200,6 +175,8 @@ in
         stress-ng
         yarn
         zeromq4
+        rsync
+        util-linux
       ];
       programs.bash.shellAliases = {
         htst = "${hptest}/bin/hptest";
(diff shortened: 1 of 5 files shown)


scripts/holoport-preflight.sh
```diff
@@ -0,0 +1,116 @@
...
+}
+
+# Evaluate the system STATE as found in the PREFLIGHT file.  Any files requiring modification or
+# removal will first be copied into the HP_PERS persistence backup directory,
+# eg. /var/lib/holoport/backup/initial/...
+case "${STATE}" in
+    initial)
+	# In the initial state (no previous holoport-preflight run), we need to remove any "default"
+	# factory identities that were shipped with the image.  This includes ZeroTier networking,
+	# SSH host keys, etc.
```
yegortimoshenko: Do I understand correctly that HoloPorts ship with cloned hard drive contents, and as such have identical SSH and ZeroTier keys?

`nixos-install` itself doesn't generate these, so unless manufacturer boots into the system and then decides to clone the hard drive, we should be fine without this particular migration.

scripts/holoport-preflight.sh
```diff
@@ -0,0 +1,116 @@
...
+    esac
+    shift
+done
+
+# Ensure that our HP_PRS persistence directory exists
+(( DRYRUN )) || [ -d "${HP_PERS}" ] || mkdir -p "${HP_PERS}"
+
+# Try to load current preflight checklist State; default to "initial"
+STATE=$( cat "${PREFLIGHT}" 2>/dev/null )
+STATE=${STATE:-initial}
```
yegortimoshenko: In order to maintain reversibility, there should be two state variables: one that describes current state version (which is what `STATE` does here) and another one that describes desired state version (that should be managed by NixOS), e.g.:

```
{ lib, ... }:

with lib;

{
  options.system.holoportos.stateVersion = mkOption {
    default = 1;
    type = types.int;
  };
}
```

yegortimoshenko: Also, I'd go with integers for state versions, starting with `0` for initial version.

yegortimoshenko:
I also think that it will become challenging and error-prone to maintain this dry run implementation. But this is something that can be addressed later on, as we have more migrations and probably after my longstanding branch merge.

modules/base.nix
```diff
@@ -266,33 +243,50 @@ in
...
+      # HoloPort Preflight checks must occur before ZeroTier and Holo / Holochain startup (or any
+      # other service which may have its configuration modified by the holoport-preflight).
+      # Furthermore, these services must be restarted whenever the Preflight service is run (eg. on
+      # automatic HoloPortOS updates).  Holochain Activation depends on running the "preflight"
+      # configuration checklist (to bring the basic HoloPort OS configuration up to standard), and
+      # then the Activation required for Holo / Holochain.  If either of these fail, the HoloPort
+      # will *not* be considered "Activated".
+      systemd.services.holoport-preflight = {
+        enable          = true;
+        before          = [ "zerotierone.service" "sshd.service" ];
```
yegortimoshenko: In time this will grow to include all system services. It would be more effective (and cause less restarts) to just `systemctl reload nginx` and `systemctl reload zerotierone` in migration that affects these particular services. This implementation will end up restarting these services on any (even whitespace) changes to `holoport-preflight.sh`.

scripts/holoport-preflight.sh
```diff
@@ -0,0 +1,116 @@
...
+case "${STATE}" in
+    initial)
+	# In the initial state (no previous holoport-preflight run), we need to remove any "default"
+	# factory identities that were shipped with the image.  This includes ZeroTier networking,
+	# SSH host keys, etc.
+	log "Clearing initial factory configurations"
+	for f in identity.public identity.secret authtoken.secret; do
+	    if [ -a    "${ZEROTIER}/${f}" ]; then
+		backup "${ZEROTIER}/${f}" "Removing default ZeroTier configuration"
+		(( DRYRUN )) || rm -f "${ZEROTIER}/${f}"
```
yegortimoshenko: More trivial issue is that say if `rm -f authtoken.secret` fails, system will be left in an inconsistent state. This can be worked around with more granular migration separation, `stat`ing files before unlinking them, etc.

The more fundamental issue, is that neither Linux nor POSIX defines any transactions over filesystems. We're fundamentally limited to pre-existing design here, but it would be nice to figure out a way to make these migrations fully atomic.

I might be missing some interesting way to make this work. Btrfs had this, until it didn't: https://github.com/torvalds/linux/commit/7a5a07a81062915c65ce27e80608b1c819b1f936

This is not just theoretical: currently, user plugging out power supply, or just outage during migration will result in inconsistent state.

scripts/holoport-preflight.sh
pjkundert: The desired state number (the CURRENT in my script) should ideally come from Nix, somehow, but I wasn't able to deduce how to access such things (eg. version numbers) from within the script.

modules/base.nix
pjkundert: Yes, perhaps that's a better plan.  I was reticent to restart/reload from within the script (because I thought that I might have to try to determine the current service state), but it would be be cleaner, I think.

scripts/holoport-preflight.sh
pjkundert: rm -f running as root should be pretty much all (every file is deleted) or nothing (no files deleted, eg. read-only fs).  It is a risk, but I think a low risk...

scripts/holoport-preflight.sh
pjkundert: I would like to try to integrate this concept.  I'll look at this tomorrow.  Thanks!

scripts/holoport-preflight.sh
samrose: I will try to find out on the mfr process, but we definitely have a case where we use the ISO and make a virtualbox, which ends up booting and snapshot the image, so this at least covers us for that case. 

When we switch to nix expression generated virtualbox image later on, this might not be an issue

scripts/holoport-preflight.sh
samrose: @pjkundert ok I will hold off on merge this morning, and I can still get it merged today to demo even if you work on it today. it won't take long to merge/build/test

samrose:
@yegortimoshenko wrote "'d like us to consider implementing a reversible migration system close to that of Active Record. If that's fine with everyone, I will do this on merge with my branch."

It seems like a good idea to me, and the "edge cases" that somehow can't be dealt with in nix channels, can be tracked then with this migration system, and if I understand you correctly, can be reversed in concert with nixos rollbacks (or advanced in terms of the record), and basically carries this needed function forward into the new holoportos branch

scripts/holoport-preflight.sh
samrose: ah, the other issue is that the mfr does boot and test the machines on the line

samrose:
@pjkundert wrote above "This affects, specifically, the ZeroTier networking. By wiping out the (default) identity.secret, we're going to lose access to the ZeroTier network, 'til the new identity is approved in ZeroTier, and entered into the cloudflare KV stores w/ the new IP address."


The machine is never authorized until holo init is run in any case

The machine is always "joined" to the specific network, but not yet authorized. Your method should achieve obtaining a fresh new id for the specific machine, then holo init will authorize it

scripts/holoport-preflight.sh
yegortimoshenko: > The desired state number (the CURRENT in my script) should ideally come from Nix, somehow, but I wasn't able to deduce how to access such things (eg. version numbers) from within the script.

Set `systemd.services.holoport-preflight.environment = { desiredStateVersion = config.system.holoportos.stateVersion; }` on Nix side, and on shell script side, `CURRENT=$desiredStateVersion` (variable names are placeholders).

pjkundert:
I feel that there is no way to reliably and safely manually trigger systemd state changes (eg. restart, reload, etc.) to other systemd services, from within a service start-up script.  It reliably deadlocks "oneshot" services, and "simple" services do not defer later systemd services -- which we need to do for the `holoport-preflight.service`.  So, I have retained the `requiredBy` for now.

When we can discern a better method, we can improve this.  It appears that even `requiredBy` does not *always* reliably restart the target services.

I feel this is complete, for now.

modules/base.nix
pjkundert: As noted below, "oneshot" services *cannot* safely and reliably manually change the state of other systemd services.  But, `requiredBy` doesn't reliably force restarting, either (this non-determinism hasn't been completely circumscribed, just observed).

scripts/holoport-preflight.sh
pjkundert: I've converted to a simple numeric state, as recommended.  I wasn't able to figure out how to use a `mkOption` to encode this state number, and reliably access this in a script.  We can look at this in later sprints, I think.

pjkundert:
Hey, @yegortimoshenko @samrose -- sorry, I've pushed another small update, to force `holoport-preflight.service` to come after `network.service` -- to ensure Ethernet MACs are available.  We'll need to pull that change into `holoportos-testnet-dev-18.09`, from `feature-holoport-preflight`, and re-trigger the build...

PR #22, opened 2019-06-24
Convert from using Ethernet MAC to root fs' Disk ID for hardware ident.

pjkundert:
o Ensure that systemd restarts holoport-preflight.service (and
  all services requiredBy it), by making it a "oneshot" service
  with RemainAfterExit = "yes".
o Scan the mount points and disk by-id to discover serialized name

### DIFF
--- a/modules/base.nix
+++ b/modules/base.nix
@@ -261,6 +261,7 @@ in
         path            = [ pkgs.rsync pkgs.utillinux ];
         serviceConfig   = {
           Type          = "oneshot";
+          RemainAfterExit= "yes";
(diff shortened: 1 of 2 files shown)


samrose:
It looks good I left one question, but merging and testing

yegortimoshenko:
Oh, that uses disk serial numbers, not FS UUIDs. Looks good!

samrose:
@yegortimoshenko the 2nd scenario is where we are producing OVA files with packer, which also snapshots the image. But this solution appears to cover both cases

PR #28, opened 2019-07-15
Hc0.0.24 alpha2

samrose:
This is an update of holo cli for new service endpoints, an update of n3h, conductor and hc cli to v0.0.24-alpha2


yegortimoshenko:
Should be safe to close. Please reopen if I'm wrong!

PR #42, opened 2019-08-22
profiles/targets: set up holochain-conductor and base hApps

samrose:
added happs, holo-cli, n3h, conductor config


profiles/targets/default.nix
```diff
@@ -50,6 +55,101 @@ in
...
+      servicelogger
+    ];
+    instances = [ 
+      {
+        agent = "host-agent";
+        dna = "happ-store";
+        id = "happ-store";
+        storage = [
+          { 
+            path = "/var/lib/holochain-conductor/happ-store";
```
yegortimoshenko: This shouldn't be hardcoded. Instead of `/var/lib/holochain-conductor`, use `config.users.holochain-conductor.home`. It will be also much easier maintain if you abstract this into `instanceConfig` similar to `dnaConfig` above.

profiles/targets/default.nix
```diff
@@ -50,6 +55,101 @@ in
...
+       ];
+       instances = [
+         { 
+          id = "hha";
+         }
+         {
+          id = "happ-store";
+         }
+         {
+          id = "holofuel";
```
yegortimoshenko: `servicelogger` is not listed here, is that by design?

profiles/targets/default.nix
```diff
@@ -50,6 +55,101 @@ in
...
+         }
+       ];
+     }
+     {
+       id = "internal-interface";
```
yegortimoshenko: What are the semantics of `internal-interface` vs `master-interface`?

overlays/overlay/n3h/default.nix
```diff
@@ -0,0 +1,8 @@
+{ pkgs, fetchzip }:
+let
+  src = fetchzip {
+    url = "https://github.com/samrose/n3h/archive/4c3541a7432b4f0ac72cb3dac83affc062f055e1.tar.gz";
```
yegortimoshenko: Also, this appears to be n3h 0.0.13 while the latest release is 0.0.20.

yegortimoshenko:
Merged in 75b588cbdecfcb18064e5a128f6947847a439c07, thank you!

PR #133, opened 2019-09-16
profiles/hydra/minion: merge into profiles/hydra

yegortimoshenko:
This fixes priority bug introduced in: 861c0314766ea4deeab9b79c37e093058582693b


profiles/hydra/master/default.nix
```diff
@@ -7,7 +7,7 @@ in
 
 {
   imports = [
-    ../minion
+    ../.
```
samrose: This is part of the idea of just turning minions into number ID machines yes?

yegortimoshenko: I'd rather say this is part of making masters also be minions. In particular, the issue that this PR fixes is Hydra stops working after #106 (because `lib.mkForce` definition has higher priority and ends up overriding Hydra SSH key that it uses to access builders).

PR #138, opened 2019-09-26
udiskie: implement udiskie in a module for hpos

samrose:


modules/system/holoportos/udiskie.nix
```diff
@@ -0,0 +1,40 @@
+{ config, pkgs, ... }:
+let
+  wrapSystemdScript = (
```
yegortimoshenko: I'm not sure what this does, could you elaborate?

modules/system/holoportos/udiskie.nix
```diff
@@ -0,0 +1,40 @@
...
+    program: ''
+      source ${config.system.build.setEnvironment}
+      exec ${program}
+    ''
+  );
+in rec {
+  environment.systemPackages = with pkgs; [
```
yegortimoshenko: For this module to work, does it really need to have these executables exposed system-wide?

modules/system/holoportos/udiskie.nix
```diff
@@ -0,0 +1,40 @@
...
+    udiskie
+    udisks2
+  ];
+  services = {
+    udisks2.enable  = true;
```
yegortimoshenko:
```suggestion
    services.udisks2.enable  = true;
```

modules/system/holoportos/udiskie.nix
```diff
@@ -0,0 +1,40 @@
...
+  };
+  systemd.user.services = {
+    udiskie = {
+      description = "Mounts disks in userspace with udisks";
+      serviceConfig = {
+        Type    = "simple";
```
yegortimoshenko:
```suggestion
        Type = "simple";
```

Modern upstream Nixpkgs identation doesn't use aligning.

modules/system/holoportos/udiskie.nix
```diff
@@ -0,0 +1,40 @@
...
+        Restart = "always";
+      };
+      script = wrapSystemdScript ''
+        ${pkgs.udiskie}/bin/udiskie \
```
yegortimoshenko: `pkgs.udiskie` should be configurable with `services.udiskie.package`.

modules/system/holoportos/udiskie.nix
yegortimoshenko: Also, `simple` is the default, so this doesn't need to be specified. See: https://www.freedesktop.org/software/systemd/man/systemd.service.html#Type=

modules/system/holoportos/udiskie.nix
```diff
@@ -0,0 +1,40 @@
...
+          -2                        \
+          --automount               \
```
yegortimoshenko: This is default, doesn't need to be specified. See udiskie man page.

modules/system/holoportos/udiskie.nix
```diff
@@ -0,0 +1,40 @@
...
+          --file-manager=""         \
```
yegortimoshenko: This and the next line should be configurable via `services.udiskie.extraFlags`.

modules/system/holoportos/udiskie.nix
```diff
@@ -0,0 +1,40 @@
...
+	  --no-notify               \
+      '';
+      wantedBy = [ "default.target" ];
```
yegortimoshenko: Usually `wantedBy` and any other options that specify systemd dependencies go on the very top of systemd unit definition.

modules/system/holoportos/udiskie.nix
```diff
@@ -0,0 +1,40 @@
...
+    };
+  };
+  users.extraUsers.user = {
```
yegortimoshenko: `extraUsers` is deprecated, this is now just `users`.

modules/system/holoportos/udiskie.nix
```diff
@@ -0,0 +1,40 @@
...
+    extraGroups = [
+      "plugdev"
+    ];
+  };
+  users.extraGroups.plugdev = { };
```
yegortimoshenko: `extraGroups` is deprecated, this is now just `groups`.

modules/system/holoportos/udiskie.nix
yegortimoshenko: Also, I don't think `udiskie` requires setting up a user. Note that this user's name is `user`, which is very prone to collisions and is not descriptive.

modules/system/holoportos/udiskie.nix
```diff
@@ -0,0 +1,40 @@
...
+	  --no-notify               \
+      '';
+      wantedBy = [ "default.target" ];
+    };
+  };
+  users.extraUsers.user = {
+    extraGroups = [
+      "plugdev"
+    ];
+  };
```
yegortimoshenko: Better syntax for all of that would be:

```suggestion
  users.users.user.extraGroups = [ "plugdev" ];
```

modules/system/holoportos/udiskie.nix
```diff
@@ -0,0 +1,40 @@
...
+  );
+in rec {
+  environment.systemPackages = with pkgs; [
+    udiskie
+    udisks2
+  ];
+  services = {
+    udisks2.enable  = true;
+  };
+  systemd.user.services = {
```
yegortimoshenko: Having as few indentation levels as possible is more readable:

```suggestion
  systemd.user.services.udiskie = {
```

modules/system/holoportos/udiskie.nix
```diff
@@ -0,0 +1,40 @@
+{ config, pkgs, ... }:
+let
```
yegortimoshenko:
```suggestion

let
```

modules/system/holoportos/udiskie.nix
```diff
@@ -0,0 +1,40 @@
...
+  environment.systemPackages = with pkgs; [
+    udiskie
+    udisks2
+  ];
+  services = {
+    udisks2.enable  = true;
+  };
+  systemd.user.services = {
+    udiskie = {
+      description = "Mounts disks in userspace with udisks";
```
yegortimoshenko:
```suggestion
```

De facto, this field is unused in most systemd units set up by NixOS. If you want to add documentation, best place to do that is NixOS options.

modules/system/holoportos/udiskie.nix
```diff
@@ -0,0 +1,40 @@
+{ config, pkgs, ... }:
+let
+  wrapSystemdScript = (
+    program: ''
+      source ${config.system.build.setEnvironment}
+      exec ${program}
+    ''
+  );
+in rec {
```
yegortimoshenko: NixOS module definitions are never recursive sets, also see indentation:

```suggestion
in

{
```

modules/system/holoportos/udiskie.nix
yegortimoshenko: Also, parentheses are not required here.

modules/system/holoportos/udiskie.nix
```diff
@@ -0,0 +1,40 @@
+{ config, pkgs, ... }:
```
yegortimoshenko: `options.services.udiskie` section is missing. That won't do, because just evaluating this file now means user will have to use udiskie.

samrose:
> I'm also unsure as to whether we need udiskie at all. Did you try https://nixos.org/nixos/options.html#autofs?

Let's try autofs before I spend time cleaning this up. I wasn't aware of it, thanks for that reference

yegortimoshenko:
Let's try that and get back to udiskie if autofs doesn't work out. Unlike udiskie, autofs seems to have been specifically designed to be used with headless systems like ours.

PR #144, opened 2019-10-10
profiles/hydra: allow superuser access to Sam Rose

yegortimoshenko:
Allows @samrose to connect to Hydra master and minion services as superuser.


samrose:
This one won't pass through totally without hydra issues being resolved fwiw @yegortimoshenko

yegortimoshenko:
@samrose You should now be able to SSH to `hydra-minion-1.holo.host`.

PR #154, opened 2019-10-28
Revert "holoportos-led-daemon: use green as idle color for internal testing"

yegortimoshenko:
This reverts commit cc104ed5d904086cf7f28f9dfb596065653f1612.

Rationale is that this was intended to verify that upgrade to new HoloPortOS from https://github.com/Holo-Host/holoportos-legacy was successful. Since we ensured that, we should revert idle LED back to its default state.


samrose:
deprecated we will resolve this in a future push to master

PR #156, opened 2019-11-07
hpos-admin: init

yegortimoshenko:
Implements Holo-Host/rfcs#2 and Holo-Host/rfcs#9. Also see: Holo-Host/holo-config#13


overlays/holo-nixpkgs/hpos-admin/hpos-admin.py
```diff
@@ -0,0 +1,65 @@
...
+
+@app.route('/v1/config', methods=['PUT'])
+def put_config():
+    state = get_state_data()
+    if request.headers['x-hpos-admin-cas'] == cas_hash(state['v1']['config']):
+        state['v1']['config'] = request.get_json(force=True)
+        with open(get_state_path() + '.tmp', 'w') as f:
+            f.write(json.dumps(state, indent=2))
+            os.rename(f.name, get_state_path())
+        subprocess.Popen(['sudo', 'nixos-rebuild', 'switch'])
```
yegortimoshenko: `nixos-rebuild` has to be mutexed, at least eventually.

overlays/holo-nixpkgs/hpos-admin/runner.py
```diff
@@ -0,0 +1,69 @@
+import logging
+import time
+
+from gevent import Greenlet, sleep
+from gevent.event import Event
+from gevent.subprocess import Popen
+
+log				= logging.getLogger( "gevent Event" )
+
+class Runner( Greenlet ):
```
yegortimoshenko: This seems to be pretty complex. `gevent.queue.Queue` would be more straightforward.

tests/holo-admin/default.nix
```diff
@@ -0,0 +1,25 @@
+{ lib, makeTest }:
+
+makeTest {
+  name = "holo-admin";
```
yegortimoshenko:
```suggestion
  name = "hpos-admin";
```

overlays/holo-nixpkgs/hpos-admin/hpos-admin.py
```diff
@@ -0,0 +1,143 @@
...
+    })
+
+@app.route('/v1/upgrade', methods=['POST'])
+def upgrade():
+    rebuild.go()
+    return '', 200
+
+
+if __name__ == '__main__':
+    logging.basicConfig( level=logging.WARNING )
```
yegortimoshenko:
```suggestion
```

This is the default.

overlays/holo-nixpkgs/hpos-admin/hpos-admin.py
```diff
@@ -0,0 +1,143 @@
...
+
+@app.route('/v1/config', methods=['GET'])
+def get_config():
+    config = get_state_data()['v1']['config']
+    return jsonify(config), 200, { 'x-hpos-admin-cas': cas_hash(config) }
+
+# APIs that require a nixos-rebuild after mutating HPOS state may trigger it using rebuild.go().
+# This is a gevent-compatible Greenlet scheduler, that ensures only one invocation runs at a time.
+
+class Runner( Greenlet ):
```
yegortimoshenko: This is too complex :(

overlays/holo-nixpkgs/hpos-admin/hpos-admin.py
```diff
@@ -0,0 +1,143 @@
...
+                    (log.warning if status.returncode else log.info)(
+                        f"{self}: Run {self.counter} Exit: {status.returncode}" )
+                except Exception as exc:
+                    log.warning(
+                        f"{self}: Run {self.counter} Exception: {exc}" )
+        finally:
+            log.info(f"{self}: Finished after {self.counter} runs, {self.trigger} triggers")
+
+rebuild = Runner( ['sudo', 'nixos-rebuild', '--upgrade', 'switch'] )
+rebuild.start()
```
yegortimoshenko: This should be inside `if __name__ == '__main__':`.

overlays/holo-nixpkgs/hpos-admin/hpos-admin.py
```diff
@@ -0,0 +1,79 @@
...
+
+def cas_hash(data):
+    dump = json.dumps(data, separators=(',', ':'), sort_keys=True)
+    return b64encode(sha512(dump.encode()).digest())
+
+
+@app.route('/v1/config', methods=['GET'])
+def get_config():
+    config = get_state_data()['v1']['config']
+    return jsonify(config), 200, { 'x-hpos-admin-cas': cas_hash(config) }
```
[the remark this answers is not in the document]
yegortimoshenko: Please also make API changes like this explicit in commit message.

overlays/holo-nixpkgs/hpos-admin/hpos-admin.py
```diff
@@ -0,0 +1,85 @@
...
+app = Flask(__name__)
+log = logging.getLogger(__name__)
+rebuild_queue = queue.PriorityQueue()
+
+
+def rebuild_worker():
+    while True:
+        (_, cmd) = rebuild_queue.get()
+        rebuild_queue.queue.clear()
+        subprocess.run(cmd)
```
pjkundert: Much simpler than Runner; however, when/if this fails it will crash hpos-admin.py; is our strategy to handle these failures with automatic restart?  Diagnosis will require looking at the `journalctl` output for the service, which is a workable approach.

overlays/holo-nixpkgs/hpos-admin/hpos-admin.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def get_state_data():
+    with open(get_state_path(), 'r') as f:
+        return json.loads(f.read())
+
+
+@app.route('/v1/config', methods=['GET'])
+def get_config():
+    return jsonify(get_state_data()['v1']['config'])
```
pjkundert: Not computing and returning the CAS (Compare-And-Swap) hash, here, *requires* the caller to compute it -- using the correct serialization (JSON with zero-space separators, sorted keys, hashed w/ SHA-512, then base-64 encoded.)  Any deviation from this knowledge will result in failure.

Including the CAS hash with every "get" allows the caller to:
1 - obtain current state w/ `x-hpos-admin-cas` value
2 - mutate the state
3 - invoke PUT `/api/v1/config` with the just-obtained `x-hpos-admin-cas` value

No knowledge of CAS hash encoding scheme required by the client.  This doesn't reduce the security or reliability of the CAS hash in any way; it still protects multiple callers from accidentally mutating the state and wiping out the earlier caller's changes.

overlays/holo-nixpkgs/hpos-admin/hpos-admin.py
pjkundert: Yes, of course! Sorry...

overlays/holo-nixpkgs/hpos-admin/hpos-admin.py
pjkundert: Along these lines; I would recommend that we augment the API to *always* include the current final config state (ie. return the results of get_config()`) and current `x-hpos-admin-cas` hash -- as the body payload of *every* PUT/PATCH request.  This costs us nothing, and provides the caller with the current state, should they wish to update UI or invoke further mutations.

pjkundert:
> This needs to be squashed or interactively rebased before merge.

Yes, I agree with a squash/rebase; I haven't been satisfied with the commit messages I've been adding, and my own clumsy attempts to solve problems that I've committed...

overlays/holo-nixpkgs/hpos-admin/hpos-admin.py
```diff
@@ -0,0 +1,85 @@
...
+        f.write(data)
+    os.rename(tmp_path, path)
+
+
+@app.route('/v1/config', methods=['PUT'])
+def put_config():
+    state = get_state_data()
+    if request.headers['x-hpos-admin-cas'] == cas_hash(state['v1']['config']):
+        state['v1']['config'] = request.get_json(force=True)
+        replace_file_contents(get_state_path(), json.dumps(state, indent=2))
```
pjkundert: There is, of course, still a race condition here.  `N` parties can `get_state_data()` ~simultaneously, compute the same `cas_hash()` -- and  then each mutate the state independently with different mutations.  The replace_file_contents is fine (uses `mstemp`, so each will be a uniquely named file) -- except that just the last of `N` parties to execute the `os.rename` will "win" the race, and create the final state.

Without additional mutual exclusion in this implementation, this can't be resolved.  It may be sufficient to:
```
# in __init__:
self.lock = Semaphore( 1 )
...
# in put_config:
with self.lock:
    state = get_state_data()
    if request.headers...:
        ...
        replace_file_contents(...)
```

overlays/holo-nixpkgs/hpos-admin/hpos-admin.py
```diff
@@ -0,0 +1,79 @@
...
+
+@app.route('/v1/config', methods=['GET'])
+def get_config():
+    config = get_state_data()['v1']['config']
+    return jsonify(config), 200, { 'x-hpos-admin-cas': cas_hash(config) }
+
+# APIs that require a nixos-rebuild after mutating HPOS state may trigger it using rebuild.go().
+# This is a gevent-compatible Greenlet scheduler, that ensures only one invocation runs at a time.
+from runner import Runner
+rebuild = Runner( ['sudo', 'nixos-rebuild', '--upgrade', 'switch'] )
```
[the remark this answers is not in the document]
pjkundert: Yes, sorry -- I deviate from PEP8 by aligning all of my mutating assignments, so all mutations in a module can be quickly visually perceived by glancing at the page of code...

pjkundert:
Implemented race condition mitigation; testing exposed issues with unprotected subprocess.run -- failure causes deadlock of queue implementation instead of exit / restart.  This was part of the "complexity" of the Runner; it handled failure of the subprocess cleanly...

yegortimoshenko:
> testing exposed issues with unprotected subprocess.run -- failure causes deadlock of queue implementation instead of exit / restart [...]

I'm not sure how that is possible, given that `check` argument defaults to `False`.

overlays/holo-nixpkgs/hpos-admin/hpos-admin.py
yegortimoshenko: CAS hash is specifically designed to be easily implementable on the client. I'm uncomfortable with client keeping track of the latest CAS hash header from logic separation and state management point of view. Please don't revert my revert of this change in this branch.

overlays/holo-nixpkgs/hpos-admin/hpos-admin.py
yegortimoshenko: Good catch!

overlays/holo-nixpkgs/hpos-admin/hpos-admin.py
```diff
@@ -0,0 +1,87 @@
...
+from hashlib import sha512
+from tempfile import mkstemp
+import json
+import logging
+import os
+
+app = Flask(__name__)
+log = logging.getLogger(__name__)
+rebuild_queue = queue.PriorityQueue()
+lock = lock.Semaphore(1)
```
yegortimoshenko: `lock.Semaphore` default is `1`. Also, `lock` here is not a descriptive global variable and clashes with import (it should be `state_lock` instead), and this section was alphasorted.

```suggestion
global_lock = lock.Semaphore()
```

overlays/holo-nixpkgs/hpos-admin/hpos-admin.py
yegortimoshenko:
```js
import stringify from 'json-stable-stringify'

const casHash = async (data) =>
  Buffer.from(await crypto.subtle.digest('SHA-512', Buffer.from(stringify(data)))).toString('base64')
```

pjkundert:
OK, the deadlock was due to gevent's subprocess.run not supporting `sudo ...` requiring a password in an expected way...  It deadlocks the process while awaiting the sudo password.

We must ensure that we set up sudo properly.

Also, to support testing via curl:
- We support POST, and partial config updates; curl doesn't support PUT
- We supply the computed CAS header from GET, so we can test using curl, without requiring complex computations to compute hashes.

overlays/holo-nixpkgs/hpos-admin/hpos-admin.py
pjkundert: In order to use the API via curl, the GET *must* return the CAS hash.  Of course, the cliennt doesn't have to use any supplied CAS hash, so it doesn't need to remember it.  But, to use via simple curl or other simplistic (eg. script based) API clients, *not* supplying the CAS via GET basically prevents the API from being usable *at* *all*.  This is not acceptable.

overlays/holo-nixpkgs/hpos-admin/hpos-admin.py
pjkundert: Yes, but its also documentary?

yegortimoshenko:
> We support POST, and partial config updates; curl doesn't support PUT

`-X PUT`

> We supply the computed CAS header from GET, so we can test using curl, without requiring complex computations to compute hashes.

Did you see https://github.com/Holo-Host/holo-nixpkgs/pull/156#discussion_r344463626?

Committed: https://github.com/Holo-Host/holo-nixpkgs/pull/156/commits/6d931fcefd58ac52fd8f4a1d72dfc2648c5ad1dc#diff-3367b4b6c55b2afa80edee8b86f60090R31
Reverted: https://github.com/Holo-Host/holo-nixpkgs/pull/156/commits/309ec8b9f8ebcba8a89ed4f00be9d61a498a82a5

Committed: https://github.com/Holo-Host/holo-nixpkgs/pull/156/commits/60bde0f3bc27a67522faf21e8ef4ecb32d16d112#diff-3367b4b6c55b2afa80edee8b86f60090R43
Reverted: https://github.com/Holo-Host/holo-nixpkgs/pull/156/commits/fd48a8cd6d16fb70ff127856a85baca92c68173b

Committed: https://github.com/Holo-Host/holo-nixpkgs/pull/156/commits/b4cf5d29926f404ec825638422e4ec321191408d#diff-3367b4b6c55b2afa80edee8b86f60090R49

yegortimoshenko:
I looked through the changes in 9ce52826b08a14f71cea8e723f5ec93dd53d1282 and ended up reverting all of them. See 16be29ca566119f4c5cc6586233d0ff700b2fdda commit message for details.

Please, follow our collaboration guidelines. For API changes, that means RFC process, for commits to this repo, it additionally means [`CONTRIBUTING.md`](https://github.com/Holo-Host/holo-nixpkgs/blob/develop/CONTRIBUTING.md). That also generally means using PR reviews to collaborate and request changes.

Please, don't push to this PR. I can't request changes on PRs I myself opened. I also can't easily review your changes.

Please, involve stakeholders when changing implementation. This latest commit, in particular, in my opinion, is almost entirely about adding things in that we shouldn't do, that can either confuse UI developers into following bad complicated state management paradigms (in reference to `x-hpos-admin-cas` in response headers), or just plain wrong (schema), or built on bad assumptions (in reference to "curl doesn't support `PUT`").

pjkundert:
OK, I was just trying to get this thing done and operational; it seems like this small bit of functionality is taking a long time to deliver.

1) The story requires that it work with `curl`; this requires that some API return the CAS, if that is what we're using to prevent simultaneous mutation -- because its just not reasonable to require all API users to be able to (or want to) compute it!  My opinion is that *every* API that modifies the state should return both A) the resultant state, and B) the CAS as a header entry (for the API user to use, if they wish).  But, to allow not even GET to return the CAS?  This is an arbitrary and difficult to recover from deficiency that prevents any simple utilization of the API by scripts.

2) Of course, the API is not "schema-less".  It is simply an undocumented schema, presently.  Allowing unbounded, arbitrary changes (up to, and including, the complete removal of the "admin" entry) is of course not correct.  Each "version" of the API may have differing restrictions on the schema, but in no case is it "schema-less".  Right now, the API simply allows a broken client to destroy the undocumented schema by writing unbounded, arbitrary garbage to the `hpos-state.json`.

3) The current implementation deadlocks.  Its easy to see; just run an API call such as `v1/upgrade` that requires sudo password, and then try another command.  Supporting logging was one method for tracking that down; documenting where to adjust that logging isn't bad practice, I don't think.  I realize we're trying to deliver, but avoiding proper logging isn't helping.  It is likely that the `gevent` monkey-patching approach to subprocess isn't handling all corner cases, and switching back to `threading` may be necessary; but proper Exception capturing and logging is a necessary first step to diagnosis, I think.


In general, its easy to deliver happy-path functionality for the 1 or 2 paths where everything works perfectly; however, almost *all* of the complexity of delivering a working production solution is in handling the 1 or 2 *hundred* ways in which the application can fail!  I was just trying to capture a few of those...  I tend to err on the side of too much error handling (and I appreciate, very much, your focus on simplicity).  But, I think we're going too far, here!  We don't detect and usefully report *any* error conditions, we don't validate *any* state changes, we don't return *any* JSON-usable `application/json` error results.  I don't think that's good.

All in all, this "simple" API doesn't follow good RESTful API practices (otherwise, we wouldn't have to use hacks like CAS to prevent bulk state mutation race conditions), and it is consuming inordinate resources to implement, and hasn't even begun to approach production code, ie. with complete error handling, input validation, comprehensive error handling and logging, useful error result reporting, etc.

I realize that much of this is probably my fault (I'm not a specialist in implementing these things, as I repeatedly have stated). So, I'll defer to the team's opinion on this.  Sorry if I've caused a problem, and for anything I've done to impede delivering this on time!

yegortimoshenko:
> The story requires that it work with `curl`; this requires that some API return the CAS, if that is what we're using to prevent simultaneous mutation -- because its just not reasonable to require all API users to be able to (or want to) compute it! 

`curl` in the story description was referring to that API works without blocking us on operational UI. It's not like we expect users to actually use curl to communicate with this API. The only consumer of `/v1/config` endpoint is HPOS Admin UI. Again, there is a reason not to do that, which I stated multiple times above. If you feel that's wrong, go through RFC process. The current state of this PR implements Holo-Host/rfcs#2 and Holo-Host/rfcs#9.

If you need to test this API, it's very easy to implement a Python HTTP client for it. It will be less than 10 lines of code. I can implement it and commit to the branch if this is the deal breaker for you.

> All in all, this "simple" API doesn't follow good RESTful API practices (otherwise, we wouldn't have to use hacks like CAS to prevent bulk state mutation race conditions) [...]

This API is REST-compliant and there is nothing about REST that implements atomic swap semantics. If you feel otherwise, point out how do you think following those practices would remove the need for CAS.

> [...] and it is consuming inordinate resources to implement

I wrote the initial version of this API in about 1 hour after reviewing your previous code. Then, I rewrote Runner into a full-on queue and that took about half an hour. This API is excruciatingly simple to implement, and the only reason it is consuming inordinate resources to implement is poor collaboration.

> So, I'll defer to the team's opinion on this.

Sure. That's why we have RFC process established to decide on things like this across the team.

> The current implementation deadlocks.

It doesn't deadlock on HoloPortOS, the platform that this service is developed for.

yegortimoshenko:
> Of course, the API is not "schema-less".

This API is schema-less. Again, look at RFC. It's up to individual services that use `hpos-state.json` to validate their own subsections of it.

overlays/holo-nixpkgs/hpos-admin/hpos-admin.py
yegortimoshenko: Just like any other default, defaults are usually left out. What would be the justification to make this default explicit?

pjkundert:
OK, then I recommend that we write a couple of quick tests in `holo-nixpkgs/tests/hpos-admin/default.nix` to generate a config, and do a GET and a PUT to confirm the APIs are working, and call it a day!  Lets get this put to bed, and move on to the next story!

samrose:
@pjkundert @yegortimoshenko I can take on writing some nixos tests for this that would allow you to move on from it, that " do a GET and a PUT to confirm the APIs are working".

For what it's worth, I really think we should invest more in understanding/respecting/contributing to/following the RFC process, and pull request processes as we've discussed them, and as many have established with known good norms/practices.

tests/hpos-admin/default.nix
```diff
@@ -0,0 +1,25 @@
+{ lib, makeTest }:
```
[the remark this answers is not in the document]
yegortimoshenko: `nix-build -A holo-nixpkgs-tests.hpos-admin`

PR #190, opened 2019-12-04
2019 12 01 hpos init logging

pjkundert:
Logging and error detection of Holo init process


pjkundert:
This tidies up a few other errors and nags that I've seen during testing.  Sorry for adding it to your existing branch, but it seems best to put it here, and get this reviewed and merged.

overlays/holo-nixpkgs/hpos-init/hpos-init.py
```diff
@@ -18,17 +20,23 @@
 wormhole send --code {} --text - < hpos-state.json
 """
 
+"""
+hpos-init -- recover and output hpos-state.json path (if successful)
```
pjkundert: This process was much complicated by the twised.reactor model; insteadd of exiting with a non-zero status code, I implemented a reliable method of transmitting indication of failure: return no 'hpos-state.json' path on stdout.

While not ideal, this is reliable.  The caller (in holo-auth-client.nix) looks for *either* a non-zero status *or* an empty value, and re-attempts the HPOS Init process in either case.  

We *cannot* successfully advance until `hpos-state.json` is recoverd, so it will loop (issuing a new magic-wormhole incantation to use) until successful.

pjkundert:
If you want to test this, checkout `2019-11-12/ci` (which includes all of these commits, in an operational package), enter a nix-shell, and hit `holoportos-shell demo`.  If you previously initialized this HoloPortOS instance, you can test most of this by removing your `/etc/hpos-state.json` (copy it to your computer, first, so you can send it via magic-wormohole).


Then, when this machine is rebooted, it'll enter the HPOS Init and Holo Auth phases; run `journalctl --follow &` to see the logging.  Try failures like using the wrong code with wormhole, or sending a non-JSON file; it'll now continue to re-attempt Holo Init.

Once you transmit the hpos-state.json, it'll attempt to Holo Auth; you'll see logging of this phase, too, and it should detect more failure modes.

pjkundert:
OK, I'm going to go ahead and merge these; there'll be hardening that needs doing, but we'll discover and address that when we start bringing up and testing HoloPortOS in earnest, today and tomorrow.

modules/services/holo-auth-client.nix
```diff
@@ -35,23 +35,38 @@ in
           zerotier-cli -j listnetworks | jq -r .[0].status
         }
 
-        sleep 10
```
yegortimoshenko: This `sleep` was intentional and is required for `wall` to work in `hpos-init`.

modules/services/holo-auth-client.nix
```diff
@@ -35,23 +35,38 @@ in
...
+        echo >&2 "Zerotier status, start: $(zerotier_status)"
 
         while [ "$(zerotier_status)" = "REQUESTING_CONFIGURATION" ]; do
-          sleep 1
+          echo >&2 "Awaiting requested ZeroTier configuration..."
+          sleep 3
```
yegortimoshenko: Would prefer not to have it bumped to 3 seconds.

modules/services/holochain-conductor.nix
```diff
@@ -28,7 +28,8 @@ in
     ];
 
     systemd.services.holochain-conductor = {
-      after = [ "network.target" ];
+      # Don't start, until keystore derived, Holo auth'ed and Zerotier online
+      after = [ "holo-auth-client.service" ];
```
yegortimoshenko: https://serverfault.com/questions/812584/in-systemd-whats-the-difference-between-after-and-requires

modules/services/holo-auth-client.nix
```diff
@@ -35,23 +35,38 @@ in
...
         done
 
-        if [ "$(zerotier_status)" = "ACCESS_DENIED" ]; then
-          export HPOS_STATE_PATH=$(hpos-init)
+        if [ "$(zerotier_status)" = "ACCESS_DENIED" ] \
+            || [ ! -s holo-keystore ] || [ ! -s holo-keystore.pub ]; then
```
yegortimoshenko: This wrongly assumes that keystore is persisted between reboots, while it shouldn't be.

modules/services/holochain-conductor.nix
pjkundert: I think this defines the relationship sufficiently; if both holo-auth-client and holochain-conductor are pulled in via some common downstream service like multi-user.target.  We should probably add a "requires" between holochain-conductor and whatever services it *needs* to have run.

modules/services/holo-auth-client.nix
pjkundert: OK, the original code didn't state this; it *only* ran derive if we were in the "ACCESS_DENIED" state.  This means  we should always run derive, regardless of the zerotier state.  Original code was defective, too...  Good catch!

pjkundert: This code needs to be re-architected, to pull out and *always* run `hpos-init` (which finds the HPOS_STATE_PATH), and `hpos-state-derive-keystore` (which derives the keys).

PR #191, opened 2019-12-04
Add some documentation and scripting to assist testers in running aarch64 VMs

pjkundert:
Figuring out how to set up and run an ARM64 VM isn't trivial; since we need to do that in order to effectively develop for the HoloPort Nano, add some documentation and scripting to help.  I've been successfully running an aarch64 Debian VM for some time using these instructions and scripting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,3 +70,10 @@ https://hydra.holo.host/job/holo-nixpkgs/master/holoportos.targets.virtualbox.x8
 Refer to [VirtualBox manual, chapter 1, section 1.15.2](https://www.virtualbox.org/manual/ch01.html#ovf-import-appliance).
 
 [nix]: https://nixos.org/nix/
+
+#### QEMU aarch64
+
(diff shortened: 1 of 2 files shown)


pjkundert:
This is just documentation, and the scripting is not used in the build or deployment of HoloPortOS, so should be pretty safe to include.  Without it, new users will find it difficult to discover how to do this...

yegortimoshenko:
If you need an Aarch64 development machine, we can provision you one on Packet.

pjkundert:
OK, i'll let it go.  But, as usual -- documentation on how to achieve this is *not in the repo*.

Documentation that doesn't exist or cannot be found is not useful.  I (nor anyone else) would be able to immediately discover: "Oh, right, I'll just `nix-build -A pkgsCross.aarch64-multiplatform.holoportos.qemu`, and then I'll use that to `... ?  ...` and then *voila*, I'll have a working, running QEMU aarch64 development platform, with all my dev tools!

PR #194, opened 2019-12-05
fixing tests

samrose:
To remove tests per @yegortimoshenko request issue also created in https://github.com/Holo-Host/holo-nixpkgs/issues


samrose:
Since this is housekeeping I am just going to merge this

PR #198, opened 2019-12-06
Fix holo auth

pjkundert:
o Await availability of root console, for `wall` communication
o Always run the holo-init and derive-keystore
o Ensure conductor requires/after holo-auth-client

### DIFF
--- a/modules/services/holo-auth-client.nix
+++ b/modules/services/holo-auth-client.nix
@@ -35,6 +35,8 @@ in
           zerotier-cli -j listnetworks | jq -r .[0].status
         }
 
+        sleep 10
+
         echo >&2 "Zerotier status, start: $(zerotier_status)"
 
         while [ "$(zerotier_status)" = "REQUESTING_CONFIGURATION" ]; do


pjkundert:
@yegortimoshenko What do you think of this improvement?  This corrects the problems, I think, and also addresses the major problem -- we are *not* always deriving the keystore!

modules/services/holo-auth-client.nix
```diff
@@ -31,41 +31,47 @@ in
...
+        echo >&2 "Deriving Holo keystore to $HPOS_STATE_PATH/holo-keystore..."
+        if ! hpos-state-derive-keystore < $HPOS_STATE_PATH > holo-keystore 2> holo-keystore.pub; then
+          echo >&2 "Failed to derive keys from $HPOS_STATE_PATH"
+          exit 1
+        fi
+        export HOLO_PUBLIC_KEY=$(cat holo-keystore.pub)
+
+        #
+        # Phase 3: Monitor Zerotier status, triggering Holo Auth if necessary
+        #
```
pjkundert: This was *all*, previously, dependent on Zerotier status being `ACCESS_DENIED` -- which would only be the case once, at the *original* time of Holo Auth.

However, we must do this *every* *boot*.  So, this is now outside the `ACCESS_DENIED` test (which only protects the `holo-auth-client` invocation, now).

modules/services/holo-auth-client.nix
```diff
@@ -18,7 +18,7 @@ in
 
   config = mkIf cfg.enable {
     systemd.services.holo-auth-client = {
-      after = [ "network.target" "zerotierone.service" ];
+      after = [ "network.target" "zerotierone.service" "systemd-logind.service" ];
```
pjkundert: No longer any need for arbitrary delay; will not start 'til `login` is running on the root console, so `wall` will work.

modules/services/holochain-conductor.nix
```diff
@@ -29,6 +29,7 @@ in
 
     systemd.services.holochain-conductor = {
       # Don't start, until keystore derived, Holo auth'ed and Zerotier online
+      requires = [ "holo-auth-client.service" ];
```
pjkundert: The conductor *cannot* start, until we've derived the keystore, so make this explicit.

modules/services/holo-auth-client.nix
```diff
@@ -31,27 +31,48 @@ in
...
+        # NOTE: This *requires* us to be able to communicate via the console, hence the
+        # require/after for systemd-logind.service
+        #
+        echo >&2 "Running HPOS init..."
+        while ! HPOS_STATE_PATH=$(hpos-init) || [ -z "$HPOS_STATE_PATH" ]; do
+          # Perhaps receiving/parsing/saving hpos-state.json failed/empty; try again...
+          echo >&2 "Failed to acheive HPOS Init; retrying..."
         done
+        export HPOS_STATE_PATH
+        echo >&2 "HPOS Init successful, with path $HPOS_STATE_PATH"
```
pjkundert: This stanza corrects the failure where, if `hpos-init` fails (eg. due to a wide variety of incorrect `hpos-state.json` failures or cryptographic failures, eg. using the wrong wormhole pass-phrase), it carries on happily.  By including this loop, we gain the capability to avoid proceeding until we actually successfully get the `hpos-state.json` file.

modules/services/holo-auth-client.nix
```diff
@@ -31,27 +31,48 @@ in
...
         mkdir -p /var/lib/holochain-conductor
         cd /var/lib/holochain-conductor
 
-        hpos-state-derive-keystore < $HPOS_STATE_PATH > holo-keystore 2> holo-keystore.pub
+        echo >&2 "Deriving Holo keystore to $HPOS_STATE_PATH/holo-keystore..."
+        if ! hpos-state-derive-keystore < $HPOS_STATE_PATH > holo-keystore 2> holo-keystore.pub; then
+          echo >&2 "Failed to derive keys from $HPOS_STATE_PATH"
+          exit 1
+        fi
         export HOLO_PUBLIC_KEY=$(cat holo-keystore.pub)
```
pjkundert: Without this loop, failure to derive the keystore are silently ignored, without logging.

modules/services/holo-auth-client.nix
```diff
@@ -31,27 +31,48 @@ in
...
+
+        while [ "$(zerotier_status)" = "REQUESTING_CONFIGURATION" ]; do
+          echo >&2 "Awaiting requested ZeroTier configuration..."
+          sleep 1
+        done
         if [ "$(zerotier_status)" = "ACCESS_DENIED" ]; then
+          echo >&2 "Authorizing Holo Auth Agent ID: $HOLO_PUBLIC_KEY..."
           exec ${cfg.package}/bin/holo-auth-client
         fi
+        echo >&2 "Zerotier status, final: $(zerotier_status)"
```
pjkundert: Only here, after `holo-init` and `derive-keystore` have completed, do we need to consider the Zerotier status.  So, we wait here 'til Zerotier gets synchronized, and then we run `holo-auth-client` if necessary.  No reason to delay based on Zerotier status earlier than this...

yegortimoshenko:
Superseded by #214.

PR #205, opened 2019-12-09
Build all DNAs using holo-nixpkgs buildDNA

pjkundert:
Repackaged all DNAs to have a top-level Cargo.toml, and to support `cargo test`.

Until we can agree that e2e testing DNAs requires a local `sim2h` service to be running on either localhost:9000 or a UNIX-domain socket, we have removed `hc test` from the `buildDNA` process.

The most violent DNA repackaging changes occurred to `happ-store`, which had a non-standard multi-level directory structure, and was difficult (if not impossible) to handle in a sensible fashion.  This was repackaged to A) move `cargo test` and `hc test` to the top level, and rename the output DNA to `happ-store.dna.json` (instead of `hApp-store.dna.json`).

### DIFF
--- a/overlays/holo-nixpkgs/build-dna/default.nix
+++ b/overlays/holo-nixpkgs/build-dna/default.nix
@@ -42,7 +42,7 @@ let
   );
 
   this = runCommand name {} ''
-    cp -Lr ${src} $out
+    cp -r ${src} $out
     chmod +w $out
     ln -s ${holochain-rust} $out/holochain-rust
   '';
@@ -50,7 +50,7 @@ let
   testDir = "${this}/test";
 
   fetchZomeDeps = name: ''
-    ln -s ${cargoToNix "${this}/zomes/${name}/code"} vendor
+    ln -s ${cargoToNix "${this}"} vendor # Expect Cargo.lock in root
   '';
 
   subDirNames = dir: attrNames
@@ -67,6 +67,7 @@ rustPlatform.buildRustPackage (
     nativeBuildInputs = nativeBuildInputs ++ [
       holochainRust.holochain-cli
       holochainRust.holochain-conductor
+      holochainRust.sim2h-server
       jq
       lld
       nodejs
@@ -83,9 +84,9 @@ rustPlatform.buildRustPackage (
   } // optionalAttrs (shell == false) {
     src = this;
 
-    preConfigure = concatStrings (map fetchZomeDeps (subDirNames "${this}/zomes"));
+    preConfigure = fetchZomeDeps "${this}"; # concatStrings (map fetchZomeDeps (subDirNames "${this}/zomes"));
 
-    RUSTFLAGS = "-C linker=lld";
+    #RUSTFLAGS = "-C linker=lld"; # Not required for Cargo.toml in root?
 
     buildPhase = ''
       runHook preBuild
@@ -98,8 +99,11 @@ rustPlatform.buildRustPackage (
     checkPhase = ''
       runHook preCheck
     '' + optionalString (pathExists (stripContext testDir)) ''
-      cp -r ${npmToNix { src = testDir; }} test/node_modules
-      hc test
+      cargo test
+      # TODO: When we are able to again perform DNA end-to-end tests, do so here, eg.:
+      #cp -r ${npmToNix { src = testDir; }} test/node_modules
+      #sim2h_server -p 9000 &
+      #hc test | ( node test/node_modules/faucet/bin/cmd.js || cat )
     '' + ''
       runHook postCheck
     '';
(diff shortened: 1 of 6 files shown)


overlays/holo-nixpkgs/build-dna/default.nix
```diff
@@ -86,9 +86,9 @@ rustPlatform.buildRustPackage (
   } // optionalAttrs (shell == false) {
     src = this;
 
-    preConfigure = concatStrings (map fetchZomeDeps (subDirNames "${this}/zomes"));
+    preConfigure = fetchZomeDeps "${this}"; # concatStrings (map fetchZomeDeps (subDirNames "${this}/zomes"));
 
-    RUSTFLAGS = "-C linker=lld";
+    #RUSTFLAGS = "-C linker=lld"; # Not required for Cargo.toml in root?
```
yegortimoshenko: This breaks Aarch64.

pjkundert: Good to know... I wonder why it breaks amd64 DNA builds presently?

overlays/holo-nixpkgs/build-dna/default.nix
```diff
@@ -43,15 +43,15 @@ let
   );
 
   this = runCommand name {} ''
-    cp -Lr ${src} $out
+    cp -r ${src} $out
```
pjkundert: No longer a necessary change; found another way to resolve issues finding DNA output artifacts, instead of using internal symbol links.

pjkundert:
This pull depends on an equivalently-named branch in all the DNA repos, which are referenced by the above commit hashes.

pjkundert:
To test, just `git checkout` the branch, and hit `nix-build -A dnaPackages`.

yegortimoshenko:
Please rebase the changes on top of `develop` in such a way that Git history makes sense. Also, while rebasing history, please see `CONTRIBUTING.md`, paticularly the commit message section.

overlays/holo-nixpkgs/build-dna/default.nix
yegortimoshenko: It doesn't, otherwise CI wouldn't pass for `develop`.

overlays/holo-nixpkgs/build-dna/default.nix
yegortimoshenko: This is a I/O performance optimization.

overlays/holo-nixpkgs/build-dna/default.nix
```diff
@@ -100,8 +101,11 @@ rustPlatform.buildRustPackage (
     checkPhase = ''
       runHook preCheck
     '' + optionalString (pathExists (stripContext testDir)) ''
-      cp -r ${npmToNix { src = testDir; }} test/node_modules
-      hc test
+      cargo test
+      # TODO: When we are able to again perform DNA end-to-end tests, do so here, eg.:
+      #cp -r ${npmToNix { src = testDir; }} test/node_modules
+      #sim2h_server -p 9000 &
+      #hc test | ( node test/node_modules/faucet/bin/cmd.js || cat )
```
yegortimoshenko: Please remove dead code. If you want to track this, it'd be best to create a GitHub issue for it. Mind that to get this fixed, in all likelihood we'll have to get Holochain to implement Unix socket support.

overlays/holo-nixpkgs/default.nix
```diff
@@ -168,6 +169,7 @@ in
   );
 
   dnaPackages = recurseIntoAttrs {
+    happ-example = callPackage ./dna-packages/happ-example {};
```
yegortimoshenko: Do we really need this in Holo Nixpkgs?

overlays/holo-nixpkgs/default.nix
```diff
@@ -116,6 +116,7 @@ in
     callPackage ./build-dna {
       inherit (llvmPackages_8) lld;
       inherit (rust.packages.nightly) rustPlatform;
+      nodejs = nodejs-12_x;
```
yegortimoshenko: Looks good, but won't be necessary if #206 gets merged.

overlays/holo-nixpkgs/default.nix
pjkundert: Not really.  But, if we don't build it (eg. to have it available in the "demo" branch), it'll go stale, I fear.  Strictly speaking, no we don't need it.

pjkundert:
OK, rebased on top of origin/develop

overlays/holo-nixpkgs/build-dna/default.nix
pjkundert: I think you're thinking `-l`, which means to use hard links; the `-L` option says to de-reference symbolic links.

yegortimoshenko: Right, thanks!

yegortimoshenko:
I am currently cleaning this up.

pjkundert:
Obsoleted by #213

PR #207, opened 2019-12-09
profiles/holoportos/demo: use sim2h for networking

pjkundert:
During early Open Alpha, we'll be using `sim2h` to provide DHT communications between nodes.

Default to the public `sim2h` server.

TODO: This resolves the "demo" branch; where else are we incorrectly assuming incorrect DHT configuration, eg. `n3h`?

### DIFF
--- a/profiles/holoportos/demo/default.nix
+++ b/profiles/holoportos/demo/default.nix
@@ -53,7 +53,8 @@ in
       dnas = map dnaConfig dnas;
       instances = map instanceConfig dnas;
       network = {
-        bootstrap_nodes = [];
+        type = "sim2h";
+        sim2h_url = "wss://sim2h.holochain.org:9000";
       };
       persistence_dir = conductorHome;
       signing_service_uri = "http://localhost:8888";


profiles/holoportos/demo/default.nix
```diff
@@ -53,7 +53,9 @@ in
       dnas = map dnaConfig dnas;
       instances = map instanceConfig dnas;
       network = {
-        bootstrap_nodes = [];
+        type = "sim2h";
+        sim2h_url = "wss://sim2h.holochain.org:9000";
+        n3h_persistence_path = "${conductorHome}/.n3h";
```
zo-el: @pjkundert - Do we need `n3h_persistence_path`?
If we are defaulting to sim2h we do not need `n3h_persistence_path`, do we?

yegortimoshenko: I don't think we do.

```suggestion
```

pjkundert: Very likely correct.

yegortimoshenko:
> TODO: This resolves the "demo" branch; where else are we incorrectly assuming incorrect DHT configuration, eg. n3h?

4d4804efa77c0bd096f579d11be170bae04214f4 has dropped n3h from Holo Nixpkgs.

yegortimoshenko:
Passed: https://hydra.holo.host/jobset/holo-nixpkgs/207

PR #208, opened 2019-12-10
Revert "2019 12 01 hpos init logging"

### LINKED ISSUE #196: Clean up after merging HPOS Init logging
See feedback on https://github.com/Holo-Host/holo-nixpkgs/pull/190.

yegortimoshenko:
Resolves #196.


profiles/hardware/qemu/default.nix
```diff
@@ -20,7 +20,7 @@ in
   virtualisation = {
     diskSize = 2048;
     graphics = false;
-    memorySize = 3072;
+    memorySize = 2048;
```
samrose: we should probably make a development profile sometime where we can modify this value, to alleviate the issue of machines crashing when people are developing etc

yegortimoshenko: Yes, good catch. I'll cherry-pick this and send as a separate PR.

PR #209, opened 2019-12-10
nixos/holo-auth-client: derive keystore on every boot

yegortimoshenko:
Depends on #208. Makes sure that we derive keystore from seed on every boot.


yegortimoshenko:
CI passed at https://hydra.holo.host/eval/645953.

PR #213, opened 2019-12-12
buildDNA: use to build all public DNAs

yegortimoshenko:
This disables tests and removes `aarch64-linux` support for the time being. We probably are not going to re-introduce it for reproducibility reasons (and instead will work on `aarch64-linux` builds in a QEMU VM). Tests will be re-enabled when in-memory DHT support is complete.

Most of Holo Nixpkgs work and all downstream work has been done by @pjkundert.


pjkundert:
Even better; new standard DNA `shell.nix` and `default.nix` are nice and easy to use to develop DNAs that require additional components for testing, etc.

PR #214, opened 2019-12-12
Correct holo-auth-client service to handle init, keystore errors, etc.

pjkundert:
o Await availability of root console, for `wall` communication
o Always run the holo-init and derive-keystore
o Ensure conductor requires/after holo-auth-client


pjkundert:
Addresses a variety of easy-to-exhibit failure modes (holo-init auth and parsing failures, keystore derivation failures), and logs progress in several areas where progress can be arbitrarily slow (eg. Zerotier auth).

Also ensures that output via `wall` will not be attempted until the root console is opened by `systemd-logind`.

pjkundert:
Several other things remain to be done, in future refinements:

We should still give the HoloPort owner the option to secure access to the loaded seed material:
  - Encrypt it with a password that can only be provided on reboot, via a Holo Auth connection
  - After loading from removable media, blacklist the media's USB ID, and prevent any further module loading

This systemd service should be split into two -- one that is triggered if the system detects that the `hpos-state.json` *is* available, early in the boot, so that it can be processed and dismounted before we achieve `multi-user.target`, and one that executes the magic-wormhole method, once `systemd-logind.service` is active?

pjkundert:
I've added yet more internal documentation of systemd service inderdependencies, and testing/logging/handling of some more failure modes I've experienced.  Start a VM, remove your /etc/hpos-state.json, and, for example:

- upload w/ an invalid key (but correct initial digit)
- upload garbage
- retry start-up and experience derive-keystore failure due to out-of-memory

These (and other failures) are now handled gracefully, with logging to indicate failure modes.

pjkundert:
Discovered that holo-auth-client.py was not validating the HTTP response code, so was declaring successful completion, even when the email address was not authorized.  Exit with failure, ensuring that the holo-auth-client.service exits with failure (and logs diagnostic details!), preventing further services from attempting to come up unsuccessfully.

pjkundert:
Now that the Qemu VM RAM size has been increased, the multiple failure modes involving holo-auth-client.service and holo-router-agent.service consuming huge amounts of RAM have been lessened.  I was seeing various other services (eg. Zerotier) being killed by the OOM while testing restarting of holo-auth-client.service, and also experienced other RAM related failures.

Once we get some mileage on the chassis, we may be able to again reduce the RAM size, but for now, it appears necessary to perform any serious testing.

pjkundert:
OK, now this just makes the existing Holo init and auth services uniformly more reliable.

PR #218, opened 2019-12-13
holo-router-agent: bump, profiles/holoportos: host HP Admin at /

yegortimoshenko:
This allows to access HPOS using a new Base36 URL and see HP Admin UI (which requires removing redirect and placing UI into root dir).


yegortimoshenko:
cc @peeech

PR #219, opened 2019-12-13
tests/holo-router: init

samrose:


yegortimoshenko:
Please move this test downstream to [Holo Router repo](https://github.com/Holo-Host/holo-router). For an example of how to set that up, see [HPOS Seed repo](https://github.com/Holo-Host/hpos-seed) (in particular, `hpos-seed-roundtrip` in `default.nix`).

PR #221, opened 2019-12-13
profiles/router-gateway: fix eval, implement HTTP -> HTTPS redirect

yegortimoshenko:
This allows Holo Router to correctly route ACME challenge so that HPOS can get a Let's Encrypt signed TLS certificate.


yegortimoshenko:
On operation side of things, `router-gateway.holo.host` no longer has `AAAA` record, because it was not working (Gateway currently doesn't handle IPv6 traffic).

PR #225, opened 2019-12-13
hpos-state: bump to 62009eeb1fe9be9bb455d3a763acb31a71cf7679

yegortimoshenko:
This updates hpos-state right to `develop` rev right before Holo-Host/hpos-config#25 was merged.


yegortimoshenko:
@zo-el This is the last version that is still `hpos-state` and not yet `hpos-config`. To introduce this as `hpos-config` into Holo Nixpkgs would mean renaming/patching a few other components, so there will be a separate PR for that.

PR #228, opened 2019-12-16
profiles/hardware/virtualbox: bump default RAM to 3GB

yegortimoshenko:
`hpos-state-derive-keystore` and `holo-router-agent` need at least 3GB RAM to reliably run.

cc @alastairong


samrose:
I'll create an issue about that question in the comment

PR #237, opened 2019-12-17
profiles/holoportos: add hylo-holo-dnas (Communities) DNA

pjkundert:
I think we should change the name to `communities` from `hylo-holo-dnas`


yegortimoshenko:
Probably doesn't make to keep https://github.com/holochain/hylo-holo-dnas/blob/201912-fix-build-dna/config.nix.

PR #246, opened 2019-12-19
201912 rust nightly

pjkundert:
Update holochain-rust to 0.0.41-alpha3
 - All DNAs updated
 - hpos-state and hp-admin-crypto updated to last `hpos-state.json` compatible version
   - These need to be updated, when we transition HPOS to `hpos-config.json`

One more pass of this to flatten all `201912-rust-nightly` branches in (in order)
 1) holochain-rust, 
 2) then holo-nixpkgs,
 3) then all hpos-state and hp-admin-crypto,
 4) and all DNAs

I've restored the Keystore::add_seed API in holochain-rust core (see pull 1999), and added some missing Cargo.toml crates to the root workspace (pull 2000), but presently these changes are in our `201912-rust-nightly` branch, which is off release-0.0.41-alpha3.


pjkundert:
Ya, needs a rebase.  I was stopped by the Holo-router - I could not get gateway compiling the rust nightly used by 0.0.41-alpha3; my attempt is in its 201912-rust-nightly branch.

pjkundert:
All the DNAs repos reference the holo-nixpkgs commit, as do hpos-config, hp-admin-crypto and holo-router repos, so those 201912-rust-nightly branches will need an update, post-rebase.  Was planning to do that tomorrow so let me know what you do! Thanks,

yegortimoshenko:
Looks like #248 includes these changes.

PR #255, opened 2019-12-22
profiles/hydra/master: use glob for holo-nixpkgs CI to see what happens

yegortimoshenko:


yegortimoshenko:
Seems to spam with way too many GitHub API requests and weirdly overrides previous check, making enforcement of CI impossible.

PR #265, opened 2019-12-28
overlays/holo-nixpkgs: bump hpos-state and hpstatus

peeech:
Update revisions of hpos-state and hpstatus to reflect changes in quickstart process.


yegortimoshenko:
This doesn't build. I will address bumps in a different PR.

PR #248, opened 2019-12-20
holochain-rust: bump to v0.0.42-alpha4

pjkundert:
DNAs and dependent packages updated to holochain-rust 0.0.41-alpha4, via their respective `release-0.0.41-alpha4-holo-nixpkgs` branches.

Run `nix-shell` and `hpos-shell`

PR #283, opened 2019-12-30
overlays/holo-nixpkgs/dna-packages: rename hylo-holo-dnas to holo-communities-dna

pjkundert:


pjkundert:
Obsolete; accomplished in #248

PR #284, opened 2019-12-31
2019 12 31 hp admin integration

peeech:
Adds fixes that allow HP Admin UI integrate with HP admin API


peeech:
changes made in this PR result in working HP Admin UI

peeech:
Moved to #295

PR #288, opened 2020-01-08
202001 hpos config schema check

pjkundert:
Runs hpos-shell w/ hpos-init and hpos-admin using hpos_config schema
 - Tested by uploading various faulty hpos-config.json files on init
 - Don't merge 'til we use the HPOS Admin UI to test hpos-admin settings

Does *not* test hpos-config.json schema inside hpos-seed server-side implementation code, because it is necessary to change the wormhole protocol flow to do it, and I couldn't determine how to accomplish this.  So, in the mean time, it uses the standard "holo-init wormhole failed" procedure, and issues a new wormhole code w/ associated exponential back-off, on schema failure during holo init.

This can be done quite easily as a future improvement.  In the mean time, both hpos-init and hpos-admin conduits to breaking the hpos-config.json file have been addressed by this pull request.

### DIFF
--- a/overlays/holo-nixpkgs/default.nix
+++ b/overlays/holo-nixpkgs/default.nix
@@ -49,8 +49,8 @@ let
   hpos-config = fetchFromGitHub {
     owner = "Holo-Host";
     repo = "hpos-config";
-    rev = "eb256e2243e08546b078c106541671fb4d4aa61d";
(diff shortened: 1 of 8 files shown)


pjkundert:
When the HPOS Admin UI is ready, lets merge this branch in, and test it.  Until then, leave it as a Draft pull request.

pjkundert:
Superceded by #289

PR #362, opened 2020-02-01
Update DNAs to use sharded sim2h servers

zo-el:
Updated pointers for:
- happ-store
- holo-hosting-app
- servicelogger 
- holofuel

### Reason for update:
Updated all DNAs to use a default sim2h server for specified DNA.

### DIFF
--- a/overlays/holo-nixpkgs/dna-packages/default.nix
+++ b/overlays/holo-nixpkgs/dna-packages/default.nix
@@ -19,8 +19,8 @@ let
   holo-hosting-app = fetchFromGitHub {
     owner = "Holo-Host";
     repo = "holo-hosting-app";
-    rev = "30b329c1ee0e4354c8ef05b8651144f01797cc17";
-    sha256 = "187w7b1gj52iypf283n10cbnd9731r0xy3bq2v8qfhdrrwp6gnb3";
+    rev = "12e57f7133c376f1bdbcb5c63135e99fbeabe771";
+    sha256 = "18isckq31j30s0krmmhb3m75dip0rqfxnp3f4000mfni0si1jwjj";
   };
 
   holo-communities-dna = fetchFromGitHub {
--- a/overlays/holo-nixpkgs/dna-packages/default.nix
+++ b/overlays/holo-nixpkgs/dna-packages/default.nix
@@ -33,8 +33,8 @@ let
   servicelogger = fetchFromGitHub {
     owner = "Holo-Host";
     repo = "servicelogger";
-    rev = "d4b411bc969e2c56436fb6c3ae5c2a2a62d26a17";
-    sha256 = "0i5a8757sikgcsrf5ppi9lbnisi5iqxh0rphkpqrd52ibpf6nfsz";
+    rev = "a464f862e757c40b8d4eadc48a7afa95c6092691";
+    sha256 = "0a2sq04jrwndpjn4fccjz4hwajfskx9jbhy974hxnm9klz01sc5r";
   };
 in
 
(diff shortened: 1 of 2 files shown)


overlays/holo-nixpkgs/dna-packages/default.nix
```diff
@@ -6,21 +6,21 @@ let
   happ-store = fetchFromGitHub {
     owner = "holochain";
     repo = "happ-store";
-    rev = "f9c5bb938376780b7e41d3234ff21baa6e04fb59";
-    sha256 = "174nhbbxcajdz8z27fhgs7r1py2ap69i8mkam2bn4pvh4skgabl4";
+    rev = "ca66b4c450c333d3ba7a3590b347754cf2c0c122";
```
peeech: This is a commit on happ-store`s `develop` while all the other commits are from `master`. Shouldn't that be unified as well?

zo-el: That is the default branch.

zo-el: Ok, I just updated the default branch in happ-store to master and updated.

zo-el:
> The default `sim2h_url` for HPOS should be `ws://public.sim2h.net:9000`. Therefore [this line](https://github.com/Holo-Host/holo-nixpkgs/blob/2020-01-31-update-dna-pkgs/profiles/logical/hpos/default.nix#L158) still needs to be updated to the new value.

Done

yegortimoshenko:
Please review rebased history: 510de08e71643b0da00e3a7cf5b358b5dd7a545f...4b58db1d4de8247a0a1dfb2738c6d7a7cd53ed9d vs 952564c74ea0e2b6fa2af3fc5e48ac2943d2a833...1c7cf4d9ba591f431279399484869928589e91c8.

In particular:
- "update" vs "updated"
- for `overlays/`, part before the colon is derivation attribute name, not path, so `dnaPackages.holofuel` not `overlays/dna-packages`
- best to break bumps into separate commits each so that it's easy to track changes and bisect/revert if necessary
- commit messages are only properly rendered in GitHub if they are shorter than ~75 chars
- for long paths, it's best to leave only the most important/best identifying parts, so `profiles/logical/hpos` is usually too long, but `logical/hpos` is much shorter

I've also changed the PR title so it's obvious what the change is for (sim2h sharding).

Check out <https://nixos.org/nixpkgs/manual/#submitting-changes-making-patches>.

profiles/logical/hpos/default.nix
```diff
@@ -155,7 +155,7 @@ in
       instances = map instanceConfig dnas;
       network = {
         type = "sim2h";
-        sim2h_url = "wss://sim2h.holochain.org:9000";
+        sim2h_url = "ws://public.sim2h.net:9000";
```
zippy: Which conductor uses this sim2h server?  I'm assuming the DNA's should specify their sim2h servers in the DNAs with the sim2h_url property, and this just is a backstop?

yegortimoshenko: My understanding is that it's a backstop, and each DNA comes with hardcoded sim2h URL property: https://github.com/holochain/happ-store/commit/74c257730201890fa0f5b503111b47fc34ffdf7a#diff-4b1eb3dc48c4e16d49db5b42298fe654R19

peeech: Yes, each of bumped DNAs have hard-coded sim2h_url in `manifest.json` under `properties.sim2h_url`. The one above is a default `sim2h_url` which is overridden whenever DNA is presenting its own. It produces the following part of conductor-config.toml:
```
[network]   
sim2h_url = "ws://public.sim2h.net:9000"
type = "sim2h"
```

PR #365, opened 2020-02-04
Integration of HP Admin UI

peeech:


peeech:
cc: @yegortimoshenko

modules/services/hp-admin-crypto-server.nix
```diff
@@ -20,6 +20,8 @@ in
     systemd.services.hp-admin-crypto-server = {
       wantedBy = [ "multi-user.target" ];
 
+      environment.RUST_LOG = "debug";
```
yegortimoshenko:
```suggestion
      environment.RUST_LOG = "info";
```

`debug` is intended only for development purposes.

overlays/holo-nixpkgs/hpos-admin/hpos-admin.py
```diff
@@ -38,7 +38,7 @@ def cas_hash(data):
     return b64encode(sha512(dump.encode()).digest()).decode()
 
 
-@app.route('/v1/config', methods=['GET'])
+@app.route('/api/v1/config', methods=['GET'])
```
yegortimoshenko: `/api` is a routing-specific implementation detail. Could you make Nginx mount `hpos-admin` with `/api` prefix? https://stackoverflow.com/a/32550251

profiles/logical/hpos/default.nix
```diff
@@ -107,31 +107,62 @@ in
...
+              return 404;
+            '';
+        };
+
+        "~ ^/holofuel(?:/.*)?$" = {
+            extraConfig = ''
+              rewrite ^/holofuel.*$ / last;
+              return 404;
+            '';
+        };
```
yegortimoshenko: This is really flaky. I'd very much prefer hash-based routing in lieu of this, but I guess we don't have much time to argue about merits of that.

profiles/logical/hpos/default.nix
```diff
@@ -107,31 +107,62 @@ in
...
+      extraConfig = ''
+        types {
+          application/javascript js;
+          application/wasm wasm;
+          image/png png;
+          image/svg+xml svg;
+          text/css css;
+          text/html html;
+        }
+      '';
```
yegortimoshenko: Again, good enough for this PR, but really, we should patch Nginx derivation to include `application/wasm wasm` MIME type.

modules/services/hp-admin-crypto-server.nix
peeech: Related idea described in #371

profiles/logical/hpos/default.nix
peeech: Don't think that we have time for this kind of redesign. At least flakiness is contained to 2 regex expressions.

profiles/logical/hpos/default.nix
peeech: I'm gonna give it a quick try, will leave as is if not successfull

overlays/holo-nixpkgs/hpos-admin/hpos-admin.py
peeech: Actually I would go even further - mount entire hpos-admin.py at `/api/v1/` as a v1 version of api, then inside API server itself have routes like: `/config`, `/settings` etc. This allows for multiple versions running simultaneously on one server!

yegortimoshenko: :+1:

profiles/logical/hpos/default.nix
peeech: Done in [c680621](https://github.com/Holo-Host/holo-nixpkgs/pull/365/commits/c680621e06a3164fb87e12848d93e67888c0ed09)

overlays/holo-nixpkgs/hpos-admin/hpos-admin.py
peeech: Done in [c680621](https://github.com/Holo-Host/holo-nixpkgs/pull/365/commits/c680621e06a3164fb87e12848d93e67888c0ed09)

profiles/logical/hpos/default.nix
peeech: Resolved in #374

PR #374, opened 2020-02-10
Patch nginx to include wasm mime type

peeech:
Original `mime.types` file in nginx does not contain an entry for application/wasm. Instead of passing `extraConfig` to nginx.conf I have patched nginx `mime.types` file.

PR #376, opened 2020-02-11
nginxStable: apply application/wasm MIME patch instead of nginx

yegortimoshenko:
See #374.

### DIFF
--- a/overlays/holo-nixpkgs/default.nix
+++ b/overlays/holo-nixpkgs/default.nix
@@ -267,7 +267,9 @@ in
 
   magic-wormhole-mailbox-server = python3Packages.callPackage ./magic-wormhole-mailbox-server {};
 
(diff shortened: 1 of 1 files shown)


peeech:
Same error as on my machine is on Hydra - check https://hydra.holo.host/build/15523/nixlog/1

overlays/holo-nixpkgs/default.nix
```diff
@@ -267,7 +267,7 @@ in
 
   magic-wormhole-mailbox-server = python3Packages.callPackage ./magic-wormhole-mailbox-server {};
 
-  nginx = previous.nginx.overrideAttrs (super: {
+  nginxStable = previous.nginxStable.overrideAttrs (super: {
```
samrose:
```suggestion
  nginx = previous.nginxStable.overrideAttrs (super: {
```

PR #378, opened 2020-02-12
2020 02 12 nginx stable fix

samrose:


samrose:
@peeech this fixes #376 by using the proper var name for `nginx` in the source tree

peeech:
Hydra passes, nginx is not actually patched. 

Build HPOS off this branch and check nginx that is up and running and you will notice that the `/conf/mime.types` file does not contain the line with `wasm`:
```
cat /nix/store/zhg107mpadyv441dx1a7way26z315ng4-nginx-1.16.1/conf/mime.types | grep wasm

```

yegortimoshenko:
This doesn't fix anything. Look at `services.nginx.package` default value.

PR #382, opened 2020-02-18
Revert all DNAs to the ones using generic sim2h_url

peeech:
Test PR for checking if sim2h communication works between DNS once they all use same sim2h server.

### DIFF
--- a/overlays/holo-nixpkgs/dna-packages/default.nix
+++ b/overlays/holo-nixpkgs/dna-packages/default.nix
@@ -6,8 +6,8 @@ let
   happ-store = fetchFromGitHub {
     owner = "holochain";
     repo = "happ-store";
-    rev = "04877c9b9219152685c20c4b5a79913c5f36f1a4";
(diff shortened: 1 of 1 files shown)


alastairong:
Do we still need to do this given the DNA upgrade issue?

peeech:
I know that transactions go through with this regression. Do they work outside of it? Will see and keep you updated @alastairong

alastairong:
Ok. I'll wait until tomorrow to switch over and reset

alastairong:
@peeech Don't worry about testing if it works on the new DNA. At PO sync we agreed to just stick with public.sim2h because not having sim2h server specified in the DNA also makes final exam testing easier.

This also buys time for a better long term solution.

peeech:
This implies that I am converting this PR to an actual PR that reverts DNA update, that way all the DNAs will use public.sim2h.net.

peeech:
I'm too lazy to do it by hand :-)

PR #391, opened 2020-02-21
hp-admin: bump to 0.10.0

zo-el:


yegortimoshenko:
This is not the correct commit message format. Please read <https://github.com/Holo-Host/holo-nixpkgs/blob/develop/CONTRIBUTING.md#commit-message-style-guide> or look at past Git history.

Instead of: `overlay/holo-nixpkgs: hp-admin bump to v0.10.0`.

It should be: `hp-admin: bump to 0.10.0`.

We actively use Git history to bisect, revert, and discover when specific changes were made. Consistent commit message format helps a lot with being able to search through the history. Hope you don't see it as needless bureaucracy.

overlays/holo-nixpkgs/default.nix
```diff
@@ -42,8 +42,8 @@ let
   hp-admin = fetchFromGitHub {
     owner = "Holo-Host";
     repo = "hp-admin";
-    rev = "2203eda5d68164645a5b044bcc68d769fc150ebf";
-    sha256 = "1lwsvh4ldjg4vzh6l4xyyghga5599ksg9f0jappp2qvzm4sxh61r";
+    rev = "543505215a882a6b5ff1a81f9a2fc941b2bce7b1";
```
yegortimoshenko: I've tagged this revision as `0.10.0` in `hp-admin` repo. If project has versions, it's better to tag them and reference them instead of revisions in Holo Nixpkgs.

PR #392, opened 2020-02-24
.github/settings.yml: allow Central and hAppy teams commit access

yegortimoshenko:
This allows anyone in Central and hAppy teams write and `develop` commit access to this repo.

cc @Holo-Host/central @Holo-Host/happy

### DIFF
--- a/.github/settings.yml
+++ b/.github/settings.yml
@@ -16,6 +16,8 @@ collaborators: []
 teams:
   - name: central
     permission: push
+  - name: happy
+    permission: push
   - name: hydra
     permission: push
 


.github/settings.yml
```diff
@@ -35,10 +37,10 @@ branches:
...
       enforce_admins: false
       restrictions:
-        users:
-          - samrose
-          - yegortimoshenko
-        teams: []
+        users: []
+        teams:
+          - central
+          - happy
```
peeech: I am sorry to be a pain but I think that lifting restrictions on `develop` is not the best idea. Both VMs and HPs can follow PRs via Hydra, so I see no reason why `develop` should not be protected / maintained by two maintainers only. It enforces proper reviewing process.

mjbrisebois: Agreed.  What I was trying to say in Mattermost was that we should be able to give them write access without compromising master and develop.  Giving hAppy push access should be enough.  That is all I have and I haven't had any issues getting changes that I need.

yegortimoshenko:
I'm pretty sure I was requested `develop` (and maybe `master`) merge access specifically, presumably in order to self-merge `hp-admin` bumps. @zo-el, could you clarify?

zo-el:
@yegortimoshenko, I agree with @peeech and @mjbrisebois, hAppy team doesn't need `merge` access to master or develop. Nor was that kind of access requested by the hAppy team.
Giving hAppy push access should be enough.

samrose:
Unless some more clarification is given I agree with @peeech for quality
control we should protect the branch

.github/settings.yml
yegortimoshenko: Addressed in the new diff.

samrose:
(this section shows that `develop` and `master` branches remain protected fwiw)

alastairong:
For my own education - what's the difference between merge and push permissions?

yegortimoshenko:
hAppy team wasn't granted rights to merge code into Holo Nixpkgs, just to (arguably) more comfortably send PRs. Review by me or Sam is still required.

samrose:
We actually both have review requirements on develop branch. And only Yegor has access to push to master at this time (which is fine with me)

alastairong:
Ok. And it looks like hAppy is also good with that so I have closed the PT story

PR #403, opened 2020-03-04
holochain-rust: version bump to v0.0.45-alpha1

peeech:
Bumps version of holochain-rust to v0.0.45-alpha1, also bumps all DNAs to a compatible version

### DIFF
--- a/overlays/holo-nixpkgs/holochain-rust/default.nix
+++ b/overlays/holo-nixpkgs/holochain-rust/default.nix
@@ -6,11 +6,11 @@ rustPlatform.buildRustPackage {
   src = fetchFromGitHub {
     owner = "holochain";
     repo = "holochain-rust";
-    rev = "v0.0.42-alpha5";
-    sha256 = "0x1vnhyq6ksjk2ci7c5kw6n1l8dwfz940sd4vzrcmzl7hhwvpd45";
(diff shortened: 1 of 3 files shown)


peeech:
cc @yegortimoshenko

overlays/holo-nixpkgs/holochain-rust/default.nix
```diff
@@ -6,11 +6,11 @@ rustPlatform.buildRustPackage {
   src = fetchFromGitHub {
     owner = "holochain";
     repo = "holochain-rust";
-    rev = "v0.0.42-alpha5";
-    sha256 = "0x1vnhyq6ksjk2ci7c5kw6n1l8dwfz940sd4vzrcmzl7hhwvpd45";
+    rev = "3a88ae4977162b4310aae4dcef1673844ec76d84";
```
zo-el: @peeech, can we use the release tag `v0.0.45-alpha1` here instead of the commit hash?

peeech: yes. I'm force-pushing

peeech: Actually no because it will change commit hash and all our dependency pyramid will collapse

PR #405, opened 2020-03-05
Test: limit holochain logging to info without state_dumps

jmday:
I noticed several of the holochain log errors are things we probably don't want to see in journalctl logs (e.g. `Aspect already being held`) so I'm trying `type = "info"` in this PR.

### DIFF
--- a/profiles/logical/hpos/default.nix
+++ b/profiles/logical/hpos/default.nix
@@ -167,6 +167,10 @@ in
         type = "sim2h";
         sim2h_url = "ws://public.sim2h.net:9000";
       };
+      logger = {
+        state_dump = false;
+        type = "info";
+      }
       persistence_dir = conductorHome;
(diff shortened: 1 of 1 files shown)


profiles/logical/hpos/sandbox/default.nix
```diff
@@ -10,6 +10,11 @@
     sim2h_url = "wss://localhost:9000";
   };
 
+  services.holochain-conductor.config.logger = {
```
yegortimoshenko: Doesn't have to be repeated twice, `hpos/sandbox` inherits from `hpos`.

yegortimoshenko:
See https://github.com/Holo-Host/holo-nixpkgs/pull/405/commits/a7a69eff3b6c9433fb90a9c742a9cc2681a59138, I squashed everything into one commit and prefixed the message with `logical/hpos`.

PR #439, opened 2020-04-14
holochain-conductor: implement config merging

filalex77:
Implements Holo-Host/rfcs#36.
Service code: https://github.com/Holo-Host/holo-update-conductor-config.
All names are subject to change :slightly_smiling_face:

### DIFF
--- a/modules/services/holochain-conductor.nix
+++ b/modules/services/holochain-conductor.nix
@@ -28,13 +28,15 @@ in
       wantedBy = [ "multi-user.target" ];
 
       preStart = ''
-      ${pkgs.hpos-config-into-keystore}/bin/hpos-config-into-keystore \
-        < $HPOS_CONFIG_PATH > /tmp/holo-keystore 2> /tmp/holo-keystore.pub
-      export HOLO_KEYSTORE_HCID=$(cat /tmp/holo-keystore.pub)
-      if [ ! -f $STATE_DIRECTORY/conductor-config.toml ]; then
-          ${pkgs.envsubst}/bin/envsubst < ${pkgs.writeTOML cfg.config} > $STATE_DIRECTORY/conductor-config.toml
-      fi
+        ${pkgs.hpos-config-into-keystore}/bin/hpos-config-into-keystore \
+          < $HPOS_CONFIG_PATH > /tmp/holo-keystore 2> /tmp/holo-keystore.pub
+        export HOLO_KEYSTORE_HCID=$(cat /tmp/holo-keystore.pub)
+        ${pkgs.envsubst}/bin/envsubst < ${pkgs.writeTOML cfg.config} \
+          | ${pkgs.holo-update-conductor-config}/bin/holo-update-conductor-config \
+          $STATE_DIRECTORY/conductor-config.toml
       '';
+
+      restartTriggers = [ cfg.config ];
 
       serviceConfig = {
         DynamicUser = true;
--- a/modules/services/holochain-conductor.nix
+++ b/modules/services/holochain-conductor.nix
@@ -36,8 +36,6 @@ in
           $STATE_DIRECTORY/conductor-config.toml
       '';
 
-      restartTriggers = [ cfg.config ];
-
       serviceConfig = {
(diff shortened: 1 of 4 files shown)


modules/services/holochain-conductor.nix
```diff
@@ -28,14 +28,16 @@ in
...
-        < $HPOS_CONFIG_PATH > /tmp/holo-keystore 2> /tmp/holo-keystore.pub
-      export HOLO_KEYSTORE_HCID=$(cat /tmp/holo-keystore.pub)
-      if [ ! -f $STATE_DIRECTORY/conductor-config.toml ]; then
-          ${pkgs.envsubst}/bin/envsubst < ${pkgs.writeTOML cfg.config} > $STATE_DIRECTORY/conductor-config.toml
-      fi
+        ${pkgs.hpos-config-into-keystore}/bin/hpos-config-into-keystore \
+          < $HPOS_CONFIG_PATH > /tmp/holo-keystore 2> /tmp/holo-keystore.pub
+        export HOLO_KEYSTORE_HCID=$(cat /tmp/holo-keystore.pub)
+        ${pkgs.envsubst}/bin/envsubst < ${pkgs.writeTOML cfg.config} \
+          | ${pkgs.holo-update-conductor-config}/bin/holo-update-conductor-config \
```
filalex77, line 35: The binary receives:
- new config (TOML) as `stdin`
- old config (file path) as `$1`
- writes output to `persistence_dir` of new config

modules/services/holochain-conductor.nix
```diff
@@ -28,14 +28,16 @@ in
...
+          $STATE_DIRECTORY/conductor-config.toml
       '';
 
+      restartTriggers = [ cfg.config ];
```
filalex77, line 39: This ensures that `conductor-config.toml` will be regenerated (and `holochain-conductor.service` restarted) on every change to `services.holochain-conductor.config`.

peeech: This code implies that rebuild of `conductor-config.toml` takes place in any of scenarios below:
- each change in `services.holochain-conductor.config`
- each restart of `holochain-conductor.service`
- reboot of HPOS
I was initially thinking that regeneration should only happen in the first scenario, but I now think that its fine to rebuild config also in other two cases, because `holo-update-conductor-config` script makes sure that this rebuild is safe for holochain's data.

filalex77:
After https://github.com/Holo-Host/holo-update-conductor-config/pull/1 is merged, `fetchFromGitHub` should be updated with the new repository commit, and this PR should be done.

modules/services/holochain-conductor.nix
```diff
@@ -28,19 +28,28 @@ in
...
+
```
peeech, line 40:
```suggestion
```
`resetTriggers` is a list of strings. Each string is representing a location in the file system that whenever this particular file gets altered the service will restart.
We do not need this functionality. I suggest removing this line.

modules/services/holochain-conductor.nix
```diff
@@ -28,19 +28,29 @@ in
...
+        export HOLO_KEYSTORE_HCID=$(cat /tmp/holo-keystore.pub)
+        ${pkgs.envsubst}/bin/envsubst < ${pkgs.writeTOML cfg.config} \
+          | ${pkgs.holo-update-conductor-config}/bin/holo-update-conductor-config \
+          $STATE_DIRECTORY/conductor-config.toml
       '';
 
       serviceConfig = {
-        DynamicUser = true;
+        User = "holochain-conductor";
+        Group = "holochain-conductor";
```
peeech, line 41: Otherwise `holochain-conductor` has no default group

peeech:
Ready to merge once Hydra's up.

zo-el:
Hi @peeech @filalex77 correct me if I am wrong. This PR is not giving the ability such that people can DNA-bump things like HoloFuel to create clean test hashspaces without having to re-deploy via hydra.

filalex77:
@zo-el In this exact code there is no such capabilities. @peeech will solve the problem you mentioned in a separate PR.

PR #440, opened 2020-04-15
usb-reset: implement via devmon

### LINKED ISSUE #188: hpos-reset: add a script that resets system state
Original implementation could be as simple as:

```
tmp=$(mktemp -d)
cp -r /etc/nixos "$tmp"
rm -r /etc /home /root /var
mv "$tmp" /etc/nixos
```

In addition to that, early systemd module should check for a flag in `hpos-config.json` or a special file on USB drive.

filalex77:
Close #188.

`services.devmon.enable` is defined by [profiles/physical/hpos/automount.nix](https://github.com/Holo-Host/holo-nixpkgs/blob/develop/profiles/physical/hpos/automount.nix#L6). If it's enabled, we add an argument `--exec-on-drive` to `ExecStart` (see [`devmon --help`](https://igurublog.wordpress.com/downloads/script-devmon/)).

This executes the script on the mountpoint after `devmon` has finished mounting the drive. This ensures proper unmounting/etc.

### DIFF
--- a/modules/system/holo-nixpkgs/usb-reset.nix
+++ b/modules/system/holo-nixpkgs/usb-reset.nix
@@ -1,0 +1,29 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+  cfg = config.system.holo-nixpkgs.usbReset;
+  checkUsbReset = pkgs.writeShellScriptBin "check-usb-reset" ''
+    if [ -f "$1/${cfg.filename}" ]; then
+      rm -f "$1/${cfg.filename}"
+      ${pkgs.hpos-reset}/bin/hpos-reset
+    fi
+  '';
+
+in {
+  options = {
+    system.holo-nixpkgs.usbReset = {
+      enable = mkEnableOption "resetting when a special USB drive is connected";
+
+      filename = mkOption {
+        description = "Name of the file to look for.";
+        type = types.str;
+      };
+    };
+  };
+
+  config = mkIf cfg.enable {
+    services.automount.execOnDrive = [ "${checkUsbReset}/bin/check-usb-reset %d" ];
+  };
+}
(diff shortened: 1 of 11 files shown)


alastairong:
Seeing an issue with the code during testing.
IMG_6310

modules/system/holo-nixpkgs/usb-reset.nix
```diff
@@ -0,0 +1,30 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+  cfg = config.system.holo-nixpkgs.usbReset;
+  devmonCfg = config.systemd.user.services.devmon;
```
filalex77, line 7: This isn't used anymore.

```suggestion
```

alastairong:
Confirmed working on my machine

overlays/holo-nixpkgs/default.nix
```diff
@@ -240,7 +240,8 @@ in
   };
 
   hpos-reset = writeShellScriptBin "hpos-reset" ''
-    rm -rf /var && reboot
+    rm -rf /var 
+    reboot
```
peeech, line 244: Great catch!

pdaoust:
@filalex77 A question for the Dev Pulse newsletter... Does this verify the integrity of the script on the USB drive before running it? With my limited understanding of Nix scripts, I didn't see anything to that effect. I can imagine some shrewd eyes inspecting the code and discovering that, and claiming that it allows an attacker to give a HoloPort owner a bad script that does arbitrary stuff (say, switching the HoloPort to a different Hydra channel). Is that possible or is it prevented by other checks?

(one thought is that the script could be part of the HPOS and the automount checker would simply check for the presence of a filename which would act as a flag to run the script.)

filalex77:
@pdaoust Your sentiment in the brackets is *exactly* what's already happening. No untrusted code from outside of HPOS can be run, and the file functions *only* as a flag. So, nothing to worry about!

pdaoust:
ah, thank you! See, I told you I don't know enough about Nix scripts ;)

PR #452, opened 2020-05-01
hpos-admin: add /hosted_happs endpoint

filalex77:


alastairong:
@peeech @filalex77 If I understand this right - an issue with renaming hApps with their hashes to enforce uniqueness is there isn't a human-readable name for HP admin API to pass to the UI. Is that right? If so can we add a hApp name somewhere?

PR #454, opened 2020-05-04
[backport] Add cache.holo.host-2 to binaryCachePublicKeys

filalex77:
Backport of new public key from #448 to `master`.

### DIFF
--- a/profiles/logical/binary-cache.nix
+++ b/profiles/logical/binary-cache.nix
@@ -4,6 +4,8 @@
   ];
 
   nix.binaryCachePublicKeys = [
+    # deprecated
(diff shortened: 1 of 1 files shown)


alastairong:
@filalex77 @peeech Since this is a hotfix, what testing should we do before deploying?

peeech:
What you wanna test here is that HPOS loads derivations from cache.holo.host. So it would be sth like:
- create test branch off master
- merge changes 
- point HPOS to this branch
- observe logs if it loads from cache or compiles system locally

alastairong:
Has this been added to any of the branches we've been testing on?
E.g. PR 440 or 444? As long as it's been added there and therefore "tested" we don't need to do any independent testing.

peeech:
I will only approve, convert to finished and merge this PR once I receive positive feedback from testers that their HoloPorts that follow #454 have passed QA

PR #459, opened 2020-05-07
profiles/hpos/development: create

filalex77:
- Add `$HPOS_DEVELOPMENT_MODE` to global environment
- Check this variable in `preStart` of `holochain-conductor.service`

### DIFF
--- a/modules/services/holochain-conductor.nix
+++ b/modules/services/holochain-conductor.nix
@@ -31,7 +31,7 @@ in
         ${pkgs.hpos-config-into-keystore}/bin/hpos-config-into-keystore \
           < $HPOS_CONFIG_PATH > /tmp/holo-keystore 2> /tmp/holo-keystore.pub
         export HOLO_KEYSTORE_HCID=$(cat /tmp/holo-keystore.pub)
-        if [[ ! -f $STATE_DIRECTORY/conductor-config.toml || $HPOS_DEVELOPMENT_MODE -ne "true" ]]; then
+        if [[ ! -f $STATE_DIRECTORY/conductor-config.toml || ! -n $HPOS_DEVELOPMENT_MODE ]]; then
           ${pkgs.envsubst}/bin/envsubst < ${pkgs.writeTOML cfg.config} \
             | ${pkgs.holo-update-conductor-config}/bin/holo-update-conductor-config \
             $STATE_DIRECTORY/conductor-config.toml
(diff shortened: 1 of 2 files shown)


peeech:
We have to merge it to master for `development` profile to be available for Holoports, right?

profiles/logical/hpos/development/default.nix
```diff
@@ -0,0 +1,5 @@
+{
+  imports = [ ../. ];
+
+  environment.variables.HPOS_DEVELOPMENT_MODE = true;
```
peeech, line 4:
```suggestion
  environment.variables.HPOS_DEVELOPMENT_MODE = "true";
```
env vars can only be of a type `string`

modules/services/holochain-conductor.nix
```diff
@@ -31,9 +31,11 @@ in
         ${pkgs.hpos-config-into-keystore}/bin/hpos-config-into-keystore \
           < $HPOS_CONFIG_PATH > /tmp/holo-keystore 2> /tmp/holo-keystore.pub
         export HOLO_KEYSTORE_HCID=$(cat /tmp/holo-keystore.pub)
-        ${pkgs.envsubst}/bin/envsubst < ${pkgs.writeTOML cfg.config} \
-          | ${pkgs.holo-update-conductor-config}/bin/holo-update-conductor-config \
-          $STATE_DIRECTORY/conductor-config.toml
+        if [ ! -f $STATE_DIRECTORY/conductor-config.toml || ! $HPOS_DEVELOPMENT_MODE ]; then
```
peeech, line 34:
```suggestion
        if [[ ! -f $STATE_DIRECTORY/conductor-config.toml || $HPOS_DEVELOPMENT_MODE -ne "true" ]]; then
```
``||`` logical operator is only avail inside `[[ ]]`

alastairong:
Why does it have to be merged to master, not just whichever branch / channel the HP is following?

filalex77:
@alastairong `master` is the branch (channel) most Holoports are following. This gives them the ability to enable developer profile/mode on top of master channel.

alastairong:
Maybe I misunderstand what we mean by `master` channel here. If we leave it in develop and with feature branches without merging to master, will HoloPorts following those branches (i.e. dev HoloPorts) be able to enable this profile?

filalex77:
If this code won't make it to `master`, people won't be able to enable development profile without first changing their Holoports to follow a channel other than `master` (e.g. `develop`), as specified in [README].

[README]: https://github.com/Holo-Host/holo-nixpkgs#changing-hpos-to-track-develop-vs-master

alastairong:
@filalex77 ok that's what I thought. I think that's fine for now i.e. we don't need to rush to get it into master since internal devs are the main users.

PR #461, opened 2020-05-11
profiles/hpos: rm symlinks in preStart of timesyncd (#460)

### LINKED ISSUE #460: systemd-timesyncd synchronisation errors may cause auto-update to fail
@filalex77 

We have had many instances where we see Network Time Synchronisation errors immediately preceding auto-upgrade.service failing to start. 

Yegor had previously said that the time sync errors could not affect auto-upgrade, but this is reported so often that there is a clear correlation worth investigating. Even if resolving time sync does not fix auto-upgrade, it will reduce the avenues to investigate.

A user on our forums did some research posted here:
> Do others see these time synchronization errors when booting up? Or you can check
> systemctl status systemd-timesyncd
> – do you get a green active (running) message? If no, I tracked that down to a NixOS issue with systemd-timesync as discussed here https://github.com/NixOS/nixpkgs/issues/31540 3
> 
> The solution discussed above worked for me: delete a symlink file so the system can properly create a directory in its place.

filalex77:
Close #460.

See https://github.com/NixOS/nixpkgs/issues/31540.

### DIFF
--- a/profiles/logical/hpos/default.nix
+++ b/profiles/logical/hpos/default.nix
@@ -247,7 +247,7 @@ in
 
   # FIXME: see https://github.com/NixOS/nixpkgs/issues/31540
   systemd.services.systemd-timesyncd.preStart = ''
-    rm -rf /var/lib/{private/,}systemd/timesyncd
+    rm -rf /var/lib/{private/,}systemd/timesync
   '';
 
   system.stateVersion = "19.09";
--- a/profiles/logical/hpos/default.nix
+++ b/profiles/logical/hpos/default.nix
@@ -245,6 +245,11 @@ in
   systemd.services.acme-default.serviceConfig.ExecStart =
     lib.mkForce "${holo-router-acme}/bin/holo-router-acme";
 
+  # FIXME: see https://github.com/NixOS/nixpkgs/issues/31540
+  systemd.services.systemd-timesyncd.preStart = ''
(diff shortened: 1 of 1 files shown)


profiles/logical/hpos/default.nix
```diff
@@ -245,6 +245,11 @@ in
   systemd.services.acme-default.serviceConfig.ExecStart =
     lib.mkForce "${holo-router-acme}/bin/holo-router-acme";
 
+  # FIXME: see https://github.com/NixOS/nixpkgs/issues/31540
+  systemd.services.systemd-timesyncd.preStart = ''
+    rm -rf /var/lib/{private/,}systemd/timesyncd
```
peeech, line 250:
```suggestion
    rm -rf /var/lib/{private/,}systemd/timesync
```
according to [issue 31540](https://github.com/NixOS/nixpkgs/issues/31540) this line should read `timesync`. Also I checked in HPOS the name of the folder created by timesyncd is `timesync`.

Unfortunatelly I do not have the way to reproduce issue #460 on my HPOS. Happy to merge and see if it changes anything.

alastairong:
@filalex77 @peeech I updated to latest develop and it still passes tests. Will merge tomorrow morning if no concerns

peeech:
Closing for now, might need to be reopen in the future once we have some solid error report

PR #462, opened 2020-05-11
bump-dna script

peeech:


overlays/holo-nixpkgs/default.nix
```diff
@@ -167,6 +176,8 @@ in
     ${remarshal}/bin/json2toml < ${writeJSON config} > $out
   '';
 
+  bump-dna = callPackage bump-dna {};
+
```
filalex77, line 180: This is not used, only `bump-dna-cli` is.

```suggestion
```

PR #464, opened 2020-05-14
[Testing 5] UI that allows multiple tx

zo-el:


alastairong:
@zo-el Per MM, hydra is not currently building new PRs. We are working on it.

PR #466, opened 2020-05-15
modules/dev-settings: init, use in profiles/development

filalex77:


filalex77:
As we are going to use `profiles.*` options in `configuration.nix` after #497, I propose to:

- remove `cfg.settings`, leaving only `cfg.enable`
- make `dev-settings.toml` to always be just a selection of attributes from `configuration.nix` set
  - currently this is only `profiles`

Thus `modules/dev-settings` is only responsible for making feature toggles made in `configuration.nix` readable by software that doesn't speak Nix but can read TOML.

peeech:
@alastairong can you organize a spike around it? I'd love to contribute but will have to download this piece of knowledge to my brain which is busy with some other stuff ATM.

alastairong:
@filalex77 @peeech How urgent is this? I would really like us to make progress with the nano before we get back to developer profile related work.

peeech:
This is a part of profiles work. So it would be healthier to finish this spike before mofving on to nanos

alastairong:
The original reason to do the profiles work was for bump-dna to work, and nothing else. That was the specific pull. If we are now able to make bump DNA work then we really cannot afford to be spending more time on the profiles.

filalex77:
Closing for now, reopening if further uses (besides `bump-dna`) emerge.

PR #469, opened 2020-05-19
Hosted Happs: Add v0.14.4-alpha1 HoloFuel for testing

alastairong:
Adding the version used by Matt for compatibility with his UI. Doesn't need to be reviewed yet. 

Need to update from develop once https://github.com/Holo-Host/holo-update-conductor-config/pull/6 is added to holo-nixpkgs

PR #471, opened 2020-05-21
Adding envoy and service logger back

alastairong:
Adding new holo-envoy from https://github.com/Holo-Host/holo-envoy

Need to update hydra config in profiles

Will need to check if we're using the right version of service logger


alastairong:
Closing as ~duplicate of #469

PR #472, opened 2020-05-26
[Testing 6]: Controlled accepting tx

zo-el:


alastairong:
@zo-el Note that this will not automatically evaluate right now. We can restart Hydra to pick it up if required but this will require it going completely offline for ~30 min. Please only request if necessary.

PR #476, opened 2020-05-28
profiles/hpos/nginx: add URI to /hosting/ proxyPass rule

peeech:
This tries to fix an error where envoy registers methods for ws-rpc in a wrong namespace, therefore those methods show to the external world as not registered


peeech:
implemented in #469

PR #497, opened 2020-06-15
modules/profiles: add development profile

filalex77:


modules/profiles/development.nix
```diff
@@ -0,0 +1,70 @@
...
+            type = types.submodule sshOptions;
+          };
+        };
+      };
+    };
+  };
+
+  config = mkIf cfg.enable (mkMerge [
+    (mkIf cfg.features.overrideConductorConfig {
+      environment.variables.HPOS_DEVELOPMENT_MODE = "true";
```
filalex77, line 58: `HPOS_DEVELOPMENT_MODE` should be renamed to `HPOS_OVERRIDE_CONDUCTOR_CONFIG`.

alastairong:
With multiple configuration options in a single profile, do we have a clear way for an API to toggle something like SSH on/off and have that persist through rebuilds and reboots?

alastairong:
@filalex77 Also this build is having infinite recursion errors apparently.

filalex77:
> a clear way for an API to toggle something

Please elaborate on what you mean by "API". This PR is meant to allow developers to add `profiles.development.*` definitions to their local `configuration.nix`.

In general, to get something like SSH working, you need to enable both the development profile and SSH feature:

```
# configuration.nix

{
  profiles.development = {
    enable = true;
    features.ssh.enable = true;
  };
}
```

Please see how backwards compatibility for profiles in `profiles/logical/hpos` is implemented here.

alastairong:
@filalex77 One of our upcoming stories is to create an endpoint for the HP admin API to toggle SSH on/off. Is that connected to this development profile? If not, are they going to play nice together?

filalex77:
@alastairong 

> are they going to play nice together?

Yes! This is exactly what we discussed with @peeech. HP admin API will create a .nix file that will be imported by `configuration.nix` and use the options defined by this PR. So yes, they will play together very nicely.

alastairong:
Ok thanks for clarifying. So if this development profile is defining a set of options that every HoloPort needs to have access to, will we be able to update all HoloPorts to include this profile? C.f. the staged profile which was added manually to configuration.nix

modules/profiles/development.nix
```diff
@@ -0,0 +1,70 @@
...
+      enable = mkEnableOption "SSH to the Holoport";
+
+      access = mkOption {
+        description = "Control SSH access";
+        type = types.submodule {
+          options = {
+            holoCentral = mkOption {
+              description = "Whether to allow Holo Central team to access the Holoport";
+              type = types.bool;
+              default = true;
```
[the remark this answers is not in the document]
filalex77: No, this means that if a user enables SSH (with `features.ssh.enable = true`), we get access by default. They can opt out by adding `features.ssh.access.holoCentral = false` to their `configuration.nix`.

filalex77:
@alastairong 

> will we be able to update all HoloPorts to include this profile?

Yes. It's the reason why I went with leaving `profiles/logical/hpos/{development,staged}` in place and forwarding their implementations to the new module. People who already added these profiles to `imports` in `configuration.nix` will still have a working configuration. All new developers should use `profiles.development.*` options directly and *not* import these files.

modules/profiles/development.nix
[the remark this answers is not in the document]
alastairong: Ok thanks :)

alastairong:
> People who already added these profiles to imports in configuration.nix will still have a working configuration. 

This is great for developers (which I know is the audience for current work) but SSH is something all our customers will need. Will users who only follow the default `profiles/target/holoport` or `holoport-plus` also be able to enable SSH this way?

filalex77:
@peeech Thanks for the review!

> updating profiles part of README

Done in c55abe7, excluding the `sandbox` profile, which is not (re)implemented by this PR.

> describing how it works user-side

Should be handled by c55abe7, but let me know if it's not enough.

> when TODOs will be gone

I thought about it, and opted to remove existing `profiles/logical/hpos/*` files that use `modules/profiles/*` in 3245191, as the reason for this PR's existence was to move from `imports` to NixOS options.

modules/profiles/README.md
```diff
@@ -0,0 +1,42 @@
...
+- `ssh.access.keys`: list of strings (SSH keys)
+  Any additional SSH (RSA, ED25519, etc.) keys that are allowed to access your Holoport.
+  Example:
+
+  ```nix
+  {
+      ssh.access.keys = [
+          "ssh-ed25519 <YOUR-ED25519-KEY-HERE>"
+      ];
+  }
```
peeech, line 41:
```suggestion
  features = {
      overrideConductorConfig = true;
      ssh = {
          enable = true;
          access = {
              holoCentral = false;
              keys = [
                  "ssh-ed25519 <YOUR-ED25519-KEY-HERE>"
              ];
          };
      };
  };
```

modules/profiles/README.md
```diff
@@ -0,0 +1,42 @@
...
+  When this feature is enabled, HPOS will no longer rewrite your `conductor-config.toml`
+  on updates, meaning that all your changes will **NOT** be overwritten.
+- `ssh.enable`: boolean, default = `false`
+  Whether to allow *any* SSH access to your Holoport.
+- `ssh.access.holoCentral`: boolean, default = `true`
+  Whether to allow Holo Central team to access your Holoport.
+  Does **not** allow access if `ssh.enable` is not `true`.
+- `ssh.access.keys`: list of strings (SSH keys)
+  Any additional SSH (RSA, ED25519, etc.) keys that are allowed to access your Holoport.
+  Example:
```
peeech, line 34:
```suggestion

    Example:
```

filalex77:
Tested on a Holoport+, everything works.

filalex77:
@alastairong Created #513.

marycamacho:
Per @alastairong This is passing tests but a GitHub bug is having it show up as failed.

PR #500, opened 2020-06-17
Add endpoint to trigger a factory reset via HP admin API

alastairong:
To call `hpos-reset` via HP Admin. Currently not passing tests

### DIFF
--- a/overlays/holo-nixpkgs/hpos-admin/hpos-admin.py
+++ b/overlays/holo-nixpkgs/hpos-admin/hpos-admin.py
@@ -104,6 +104,14 @@ def upgrade():
     return '', 200
 
 
+@app.route('/reset', methods=['POST'])
+def reset():
+    try:
+        subprocess.run(['hpos-reset'], check=True)
+    except CalledProcessError:
+        return '', 400
+
+
 def unix_socket(path):
     sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
     if os.path.exists(path):
--- a/overlays/holo-nixpkgs/hpos-admin/hpos-admin.py
+++ b/overlays/holo-nixpkgs/hpos-admin/hpos-admin.py
@@ -109,7 +109,7 @@ def reset():
     try:
         subprocess.run(['hpos-reset'], check=True)
     except CalledProcessError:
-        return '', 400
+        return '', 500
 
 
 def unix_socket(path):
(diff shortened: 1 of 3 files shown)


overlays/holo-nixpkgs/hpos-admin/hpos-admin.py
```diff
@@ -104,6 +104,14 @@ def upgrade():
     return '', 200
 
 
+@app.route('/reset', methods=['POST'])
+def reset():
+    try:
+        subprocess.run(['hpos-reset'], check=True)
+    except CalledProcessError:
+        return '', 400
```
peeech, line 112: 400 means `The server could not understand the request due to invalid syntax.`. The range of errors 4xx is reserved for errors coming from client. You probably want to return [500](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/500).
In case of successful execution of `hpos-reset` there's gonna be a race condition: `hpos-reset` is a two-liner that in the second line of code reboots HPOS. I think that python sub-process is gonna finish first and manage to send a 200 response before the reboot target is reached. But I am not sure, so it might leave UI in an unfinished request state.

alastairong: Once the first line of `hpos-reset` executes, will the UI still be able to maintain a websocket connection? Won't hp-admin-crypto auth fail?

If so, perhaps the simplest way is to assume that disconnection = success case and handle it through clear instructions / messaging.

alastairong: Fixed error code to 500 in `4a109dd`

alastairong:
I have removed tests for now because the test engine cannot handle a VM reboot and terminates. Will revisit once we migrate to 20.03 with Python test engine

alastairong:
Adding issue: https://github.com/Holo-Host/holo-nixpkgs/issues/503

alastairong:
Screenshot 2020-06-18 at 11 27 53

Screenshot with proof that that factory reset does execute. Ghetto tests FTW! XD

overlays/holo-nixpkgs/hpos-admin/hpos-admin.py
peeech: The endpoint does what it's supposed to do - resets HPOS. The question still remains - how is UI going to handle this. After reboot HPOS wakes up without any identity, which means FQDN that HP Admin is trying to reach is no longer existing. But that's the UI/UX issue.

alastairong:
@filalex77 Can you merge this please?

PR #501, opened 2020-06-17
profiles/hpos: add descriptive fields

alastairong:
Adds fields for:
- title
- publisher
- publish date
- release version

To hosted hApps in conductor-config.toml so that HP admin can access them
These fields are supposed to be arbitrary strings with no function


alastairong:
@peeech You will need to merge this. I don't have admin access.

PR #502, opened 2020-06-17
HP Admin return number of hosted instances per happ

alastairong:
Updates the `/hosted_happs` endpoint to also provide the number of hosted instances for each hosted_happ ID.

### DIFF
--- a/overlays/holo-nixpkgs/hpos-admin-client/hpos-admin-client.py
+++ b/overlays/holo-nixpkgs/hpos-admin-client/hpos-admin-client.py
@@ -49,7 +49,13 @@ def put_settings(ctx, k, v):
 @cli.command(help='Get HoloPortOS status data')
 @click.pass_context
 def get_status(ctx):
-    print(request(params, 'GET', '/status').json())
+    print(request(ctx, 'GET', '/status').json())
+
+
+@cli.command(help='Get info on happs currently hosted')
+@click.pass_context
+def get_hosted_happs(ctx):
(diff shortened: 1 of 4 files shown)


alastairong:
@peeech 

@zo-el asked if, since we were exposing conductor-config data via API, could we show the number of instances / users that way instead of using service logger to do so.

This PR is to do that, but I would like to confirm with you whether these approaches give identical results - I.e. are # of instances in conductor-config.toml == # source chains

peeech:
Each instance creates its own source chain stored in `instances.storage` file. So yes - that equality holds.

alastairong:
@filalex77 I implemented a string equality test as I wasn't able to figure out how to parse JSON properly in PERL. While it's technically an unordered dict I noticed that the API consistently returns the JSON in alphabetical order  across multiple machines.

overlays/holo-nixpkgs/hpos-admin-client/hpos-admin-client.py
```diff
@@ -49,7 +49,13 @@ def put_settings(ctx, k, v):
 @cli.command(help='Get HoloPortOS status data')
 @click.pass_context
 def get_status(ctx):
-    print(request(params, 'GET', '/status').json())
```
alastairong, line 52: This change isn't technically part of this PR but I noticed that params is undeclared and should be ctx instead.

alastairong:
@filalex77 Can you merge this in please?

PR #509, opened 2020-06-23
Add wormhole error handling to fix customer bug

alastairong:
Customer mrcat in China reports that their HoloPort cannot connect to our network. Screenshot suggests this is due to an unhandled error with Wormhole, which is a currently unused service set up for future VMs and Cloudports.

This is a basic error handler to address the issue for HP owners - we will probably need something more robust when Cloudports are operating.

图片

Note that this nixOS project also had a very similar / the same issue and it may be caused with an unhandled error in the Twisted library.
https://github.com/Tribler/tribler/issues/3440#issuecomment-400987017


alastairong:
What are your confidence levels that my change would actually address the issue in the screenshot? Because I haven't been able to reproduce or test this...

alastairong:
@zo-el @peeech Can we add this to our next test so we can merge it to develop after. It's a very minor error handling fix.

PR #510, opened 2020-06-23
overlays/holo-nixpkgs: disable QEMU/VBox images

filalex77:
As noted in the comment, these images are responsible for long Hydra build times, and are rarely used (or not at all).


alastairong:
To be honest it's not a particular pain point for our testers, but it is noticeable so let's remove it for now

alastairong:
@peeech @filalex77 I saw this written in the README. Is this still true?

> If you have Nix installed, checkout the repo, enter nix-shell and then run hpos-shell. That will launch a HoloPortOS VM against current state of your local checkout, which is useful for iterative development.

peeech:
Closing for now as this is not a priority ATM

PR #518, opened 2020-07-14
staging candidate 

peeech:
functionally replaces #515


peeech:
BTW last update to README.md is just to trigger Hydra re-evaluation so that github receives build status correctly :-)

PR #519, opened 2020-07-15
profiles/hpos: add git to systemPackages

filalex77:


peeech:
What is this PR trying to fix?

alastairong:
@peeech we occasionally see update issues related to git missing. Jarod and one of our testers reported it during this test cycle, and Oleksii saw it before.

It might also be a cause of our auto-upgrade issues (auto-upgrade doesn't log the error in systemctl status, but nixos-rebuild switch reports it)

It seems like the easiest thing to do would just be to add it as a global package.

peeech:
If we were able to identify which process is reporting this error then adding `git` to this particular processes environment would make sense more, wouldn't it? But I am not opposing proposed solution because git is useful as a system-wide package no matter what :-)

alastairong:
@peeech Yes that would be ideal. Also - we're seeing loud fan noises on some holoports, so maybe we can't add git as a global package

peeech:
My holoport was going bananas with fan, as well. It turned out it was holochain spinning at 700% CPU. I think that this might be a source of fan-spinning.

alastairong:
Oh crap I accidentally merged a bunch of stuff into this...@peeech @filalex77 

How do I revert? Or is it easier just to make a new PR?

peeech:
git revert ef4519d -m 1

alastairong:
@peeech That creates it's own mess. Any issue if I close this and open a new PR? It's a 3 char 1 line change...

peeech:
yeah you're right, test-13 was its own mess. You can close this PR, cherry pick `ab502b0` into new PR and then delete `origin/2020-07-15/add-git-to-hpos`

PR #516, opened 2020-07-13
Add hpos-update script for testers

alastairong:
This script is for making testing easier by simpler channel switching and update success feedback

PR #521, opened 2020-07-15
profiles/hpos: add git to systemPackages

alastairong:
We occasionally see update issues related to git missing. Jarod and one of our testers reported it during this test cycle, and Oleksii saw it before.

It might also be a cause of our auto-upgrade issues (auto-upgrade doesn't log the error in systemctl status, but nixos-rebuild switch reports it)

It seems like the easiest thing to do would just be to add it as a global package.

This is a duplicate of #519 which I screwed up :/

### DIFF
--- a/overlays/holo-nixpkgs/default.nix
+++ b/overlays/holo-nixpkgs/default.nix
@@ -257,6 +257,8 @@ in
     reboot
   '';
 
+  inherit (callPackage ./hpos-update {}) hpos-update-cli;
+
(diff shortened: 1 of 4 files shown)


alastairong:
@filalex77 @peeech Shall we update and merge this in?

filalex77:
@alastairong I think this change was made in 3032df5def026489f5facda65600da1d852731fa as part of #516.

PR #529, opened 2020-07-21
Allow HP admin to toggle HPOS features

filalex77:
Routes added to HP Admin:
- `GET /profiles`: get the whole TOML file generated by HP Admin
- `GET /profiles/<profile>/features/<feature>`: check if a feature is enabled
  - returns `{"enabled": bool}`
- `PUT /profiles/<profile>/features/<feature>`: enable a feature
- `DELETE /profiles/<profile>/features/<feature>`: disable a feature

### DIFF
--- a/modules/system/holo-nixpkgs/auto-upgrade.nix
+++ b/modules/system/holo-nixpkgs/auto-upgrade.nix
@@ -16,15 +16,7 @@ in
   };
 
   config = mkIf cfg.enable {
-    systemd.services.holo-nixpkgs-auto-upgrade = {
-      serviceConfig.Type = "oneshot";
-      unitConfig.X-StopOnRemoval = false;
-      restartIfChanged = false;
-
-      environment = config.nix.envVars // {
-        inherit (config.environment.sessionVariables) NIX_PATH;
-        HOME = "/root";
-      } // config.networking.proxy.envVars;
+    systemd.services.holo-nixpkgs-auto-upgrade = config.systemd.services.nixos-rebuild // {
 
       path = with pkgs; [ config.nix.package git gzip gnutar xz curl jq perl ];
 
--- a/modules/system/holo-nixpkgs/auto-upgrade.nix
+++ b/modules/system/holo-nixpkgs/auto-upgrade.nix
@@ -17,6 +17,7 @@ in
 
   config = mkIf cfg.enable {
     systemd.services.holo-nixpkgs-auto-upgrade = config.systemd.services.nixos-rebuild // {
+      after = [ "network.target" ];
 
       path = with pkgs; [ config.nix.package git gzip gnutar xz curl jq perl ];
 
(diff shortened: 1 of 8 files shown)


alastairong:
@filalex77 Has this been added in a way that means these options will be automatically enabled for every HoloPort **without** requiring any of our users to manually access `configuration.nix`?

alastairong:
@filalex77 I also forgot that per our definition of done, you will need to update the RFC. Feel free to add it to this https://github.com/Holo-Host/rfcs/pull/49

And then we should merge that documentation.

peeech:
Temporarily closing - my suspicion is that a kind of errors that this PR gives on Hydra is a source of our problems with hanging Hydra.

filalex77:
Fixed infinite recursion errors in 0774300444b89aae318e222f9d54cd8fac37db83, and allowed for `/etc/nixos/hp-admin-features.toml` to not exist in 3c01df628b9d1c83ac8e30108895489399d10fd2.

Reopening this PR, as it should be safe for Hydra evaluation now.

alastairong:
@filalex77 Fails to build. Looks like a syntax error in the tests

alastairong:
@filalex77 I pushed up a fix for the perl syntax. Still failing though

overlays/holo-nixpkgs/hpos-admin/hpos-admin.py
```diff
@@ -213,7 +217,7 @@ def status():
 
 @app.route('/upgrade', methods=['POST'])
 def upgrade():
-    rebuild(priority=1, args=['--upgrade'])
+    rebuild_upgrade(priority=1)
```
alastairong, line 220: Does this mean that /upgrade endpoint will work?

modules/system/holo-nixpkgs/auto-upgrade.nix
```diff
@@ -16,15 +16,8 @@ in
   };
 
   config = mkIf cfg.enable {
-    systemd.services.holo-nixpkgs-auto-upgrade = {
```
alastairong, line 19: It looks like we're just moving the auto-upgrade itself to a separate nixos-rebuild service, and then calling it here. Can you explain a little more about what and why?

filalex77: Thanks for pointing out. I lost a changed `script` which calls `nix-channel --update` before rebuilding. That's what code reviews are for 😅 👀

overlays/holo-nixpkgs/hpos-admin/hpos-admin.py
filalex77: Not yet, I need to resolve some newly appeared issues beforehand. After that, it should work.

modules/system/holo-nixpkgs/auto-upgrade.nix
filalex77: Actually didn't lose it, misread the diff.

So this change is for sharing environment variables that Nix needs for `nixos-rebuild` between the two services, using `nixos-rebuild.service` as a base.

filalex77:
Closing in favor of #555.

PR #538, opened 2020-07-27
overlays/hpos-admin: Change upgrade endpoint mechanism

alastairong:
Changing hpos-admin /upgrade endpoint to use `hpos-update` rather than `nixos-rebuild switch` so that the new local revision is saved in .nix-revision file.


alastairong:
Closing as superceded by the work on #529

PR #542, opened 2020-07-29
profiles/holo: add trusted people to Hydra access

filalex77:
Add @evangineer, @JettTech, @zo-el.
Keys taken from [modules/profiles/holocentral_keys](https://github.com/Holo-Host/holo-nixpkgs/blob/develop/modules/profiles/holocentral_keys).


evangineer:
@filalex77 It seems that @JettTech is being listed as having the identical public key to me.  

Clearly that's not correct.

PR #550, opened 2020-08-06
[WIP]: Hosted happs (Testfuel)

zo-el:


profiles/logical/hpos/default.nix
```diff
@@ -21,13 +21,14 @@ let
   '';
 
   conductorHome = "/var/lib/holochain-conductor";
+  holofuelHhaHash = "QmWvWtw1P2nduQqk8uKpfhAeUd72bSU2gFH7TLrQm1Bbfw";
```
zo-el, line 24: Should we be auto generating this?

alastairong: Yes and it needs to be exactly the same as what hpos is registering itself in KV store.

On boot, HPOS will always register itself for every happ that is hosted. So this can't be arbitrary otherwise there will be conflicts.

@peeech - we need to align on this

PR #551, opened 2020-08-12
HPOS update handles cache result scenario

alastairong:
1) Confirmed that this "does no harm" - `hpos-update` works as usual in the success case
2) Confirmed that this fixes standard errors leading to false positive. When update fails, `get-status` will show the correct (pre-update) holo-nixpkgs hash
3) Confirmed that script captures the correct output to `grep` for cached results
4) CAN NOT reproduce actual cached result error


alastairong:
I will be on vacation so please feel free to merge this.

PR #575, opened 2020-09-28
[Testing] Hosted happ release (testfuel.holo.host) 

zo-el:


modules/services/holo-envoy.nix
```diff
@@ -23,6 +23,7 @@ in
 
       serviceConfig = {
         Environment = "LOG_LEVEL=silly";
+        ExecStartPre = "/run/current-system/sw/bin/sleep 10";
```
mjbrisebois, line 26: I could not figure out how to use the proper nix incantation for calling `sleep`

cc @peeech @Br1ght0ne

zo-el: @mjbrisebois - We should just use the work @peeech did in this PR: https://github.com/Holo-Host/holo-nixpkgs/pull/579

PR #577, opened 2020-09-30
[hydra] NixOS 20.03

Br1ght0ne:
Leftover tasks:
- [ ] review and fix tests
- [ ] remove `hydra-2003.holo.host` when replacing Hydra 19.09 behind `hydra.holo.host`
- [ ] rebase this PR onto `master`/`develop`
  - [x] #593 is rebased onto `2020-10-01-switch-to-RSM`


Br1ght0ne:
Ran `git rebase -i hydra` and fixed merge conflicts. Re-reviewed the PR, but I'd appreciate a look @peeech @alastairong

modules/default.nix
```diff
@@ -4,7 +4,6 @@
   imports = [
     ./services/aorura-emu.nix
     ./services/automount.nix
-    ./services/dnscrypt-proxy2.nix
```
peeech, line 7: I wanna make sure that `holo-router-gateway` is working properly with this change before we merge this PR into the branch that `holo-router-gateway` is pulling from.

profiles/logical/holo/hydra/master/default.nix
```diff
@@ -42,15 +42,25 @@ in
   nix.distributedBuilds = true;
 
   nix.extraOptions = ''
+    allowed-uris = https://github.com
     builders-use-substitutes = true
+    restrict-eval = false
```
peeech, line 47: I think I'd like to better understand why are we switching from `restrict-eval = false` to `restrict-eval = true`

profiles/logical/holo/hydra/master/default.nix
```diff
@@ -42,15 +42,25 @@ in
...
   '';
+  
+  security.acme = {
+    acceptTerms = true;
+    # REVIEW: maybe a dedicated email for Hydra?
+    email = "oleksii.filonenko@holo.host";
```
peeech, line 53: Yes. So which email shall we use here @alastairong ?

profiles/logical/holo/hydra/master/default.nix
```diff
@@ -59,7 +69,7 @@ in
       binary_cache_public_uri = https://cache.holo.host
       evaluator_max_heap_size = ${toString (4 * 1024 * 1024 * 1024)}
       log_prefix = https://cache.holo.host/
-      max_concurrent_evals = 3
+      max_concurrent_evals = 1
```
peeech, line 72: Are we staying on `max_concurrent_evals = 1` or shall we bump up to something more. I am not 100% sure what would be an advantage for higher number. We are limited by CPU no matter what.

profiles/logical/hpos/default.nix
```diff
@@ -90,6 +90,11 @@ in
     options = "--delete-older-than 7d";
   };
 
+  security.acme = {
+    acceptTerms = true;
+    email = "acme@holo.host";
```
peeech, line 95: Same question as above @alastairong . Is that the right email address to use?

peeech:
>     * fix `hpos-admin` test
>     * rebase this PR onto `master`/`develop`

The content of this PR is needed for a release on Holochain RSM. So my suggestion is to rebase this PR to `2020-10-01-switch-t0-RSM`. This also implies that tests will require much more rewrite. I would be fine disabling tests in this PR and dealing with them in a separate PR.

profiles/logical/holo/hydra/master/default.nix
Br1ght0ne: I think 3-4 (limited by CPU cores) is reasonable.

profiles/logical/holo/hydra/master/default.nix
Br1ght0ne: Hydra doesn't care about `restrict-eval`, which is Nix-specific (e.g. `nix-build`). I think we should leave it set to `false` to avoid any breakage when manually evaluating parts of `holo-nixpkgs`.

profiles/logical/holo/hydra/master/default.nix
alastairong: Let's use @peeech for now ;)

alastairong: Or you can use mine

PR #580, opened 2020-10-02
holochain rsm part 1

peeech:
Ground work for switching from Holocahin Redux to Holochain RSM
- removes holochain-redux
- removes components used exclusively by redux
- adds Holochain RSM derivation
- adds Lair Keystore derivation
- disables test


profiles/logical/hpos/default.nix
```diff
@@ -21,56 +21,6 @@ let
   '';
 
   conductorHome = "/var/lib/holochain-conductor";
```
zo-el, line 23: Should we change the home directory to `/var/lib/holochain`?

profiles/logical/hpos/default.nix
```diff
@@ -185,21 +135,7 @@ in
   services.holochain-conductor = {
```
zo-el, line 135: Should we rename this service `holochain-conductor` to just `holochain`?

peeech: I more than agree that `holochain-conductor` should be renamed to `holochain`. The reason I would like to leave it as is for now is that it's easier to identify legacy elements in the code by the name `conductor`, therefore it will be easier to remove them once the switch to RSM is finished.

profiles/logical/hpos/default.nix
alastairong: That is probably more consistent, but if so be aware that `holo-router-agent` expects the existence of that folder, so we need to make a minor update to `holo-router-agent` too

alastairong (approved):

profiles/logical/hpos/default.nix
peeech: oh yeah there's a bunch of places that not only have this path hardcoded, but also require certain structure of this file. This structure has changed with RSM so there will be a lot of work to make sure things work in an expected way

PR #582, opened 2020-10-05
Switch to Holochain RSM

peeech:


modules/services/holochain.nix
```diff
@@ -0,0 +1,57 @@
...
+    };
+
+    package = mkOption {
+      default = pkgs.holochain;
+      type = types.package;
+    };
+
+    working-directory = mkOption {
+      default = "";
+    };
```
Br1ght0ne, line 24: What happens when this is not set (e.g. `""`)? Is behaviour of `systemd.services.<name>.serviceConfig.WorkingDirectory` well-defined? I have some doubt about this being weakly-typed.

```suggestion
    working-directory = mkOption {
      type = types.path;
    };
```

`types.path` makes more sense (and add some basic validations) over `types.str`, and omitting `default` means that it's a hard error if we, for example, forget to set it in `profiles/logical/hpos`.

Br1ght0ne (commented):
Skim-level review: LGTM, but see comment.

modules/services/holochain.nix
peeech: Yes, I like this catch!!

Br1ght0ne (approved):

PR #583, opened 2020-10-05
Configure holochain service

peeech:
This pull request is aiming to create holochain service that:
- starts without errors
- communicates with keystore-lair

### DIFF
--- a/modules/services/hpos-admin.nix
+++ b/modules/services/hpos-admin.nix
@@ -25,6 +25,7 @@ in
     systemd.services.hpos-admin-socket-setup.script = ''
       chgrp hpos-admin-users /run/hpos-admin.sock
       chmod g+w /run/hpos-admin.sock
+      rm -rf /var/lib/holochain-conductor
     '';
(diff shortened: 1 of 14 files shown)


modules/services/hpos-admin.nix
```diff
@@ -25,6 +25,7 @@ in
     systemd.services.hpos-admin-socket-setup.script = ''
       chgrp hpos-admin-users /run/hpos-admin.sock
       chmod g+w /run/hpos-admin.sock
+      rm -rf /var/lib/holochain-conductor
```
zo-el, line 28: Should we write a note here to remind us why we need to remove the old file?

zo-el (approved):
LGTM. Just left one comment.

modules/services/hpos-admin.nix
peeech: Yeah... the note left here will not persist in our collective knowledge. I will find better place to document this cleanup script.

PR #585, opened 2020-10-06
holo-hosting release 

zo-el:
This is the work that was dev tested on #575

### DIFF
--- a/profiles/logical/hpos/default.nix
+++ b/profiles/logical/hpos/default.nix
@@ -175,16 +175,6 @@ in
           proxyPass = "http://127.0.0.1:4656";
           proxyWebsockets = true;
         };
-
-        "/hc/master/" = {
-          proxyPass = "http://127.0.0.1:42211/";
-          proxyWebsockets = true;
-        };
-
-        "/hc/admin/" = {
-          proxyPass = "http://127.0.0.1:42233/";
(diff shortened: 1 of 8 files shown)


zippy (approved):

profiles/logical/hpos/default.nix
```diff
@@ -170,6 +175,16 @@ in
...
+
+        "/hc/master/" = {
+          proxyPass = "http://127.0.0.1:42211/";
+          proxyWebsockets = true;
+        };
+
+        "/hc/admin/" = {
+          proxyPass = "http://127.0.0.1:42233/";
+          proxyWebsockets = true;
+        };
```
peeech, line 187: Quick question - which service is making use of those endpoints?

zo-el: `/hc/master/`: This port is used by tryorama when it tries to connect to a holoport to get consistency information.
`/hc/admin`: I am not sure.

@mjbrisebois is there any use of those endpoints in the hosted pipeline?

mjbrisebois: No, these are not supposed to stay in hpos.  They are needed to inspect conductor during debugging from outside the OS, but they are security holes that cannot stay there.

peeech (approved):
Now it looks great!

PR #588, opened 2020-10-13
profiles/hpos: replace simp_le with lego

Br1ght0ne:
This is 19.09 compatible.

### DIFF
--- a/profiles/logical/hpos/default.nix
+++ b/profiles/logical/hpos/default.nix
@@ -8,16 +8,12 @@ let
     until $(${curl}/bin/curl --fail --head --insecure --max-time 10 --output /dev/null --silent "https://$base36_id.holohost.net"); do
       sleep 5
     done
-    exec ${simp_le}/bin/simp_le \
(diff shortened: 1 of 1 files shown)


alastairong:
@Br1ght0ne In our pinned version of 19.09 there is no option for `config.security.acme.email`. That was added later (I think in 20.03). Instead, there is a per-certificate option to set email address.

Rather than change the code to be per-certificate I figured we could hardcode it for now and switch it to `config.security.acme.email` when we upgrade to 20.03

alastairong:
This PR breaks the holo-router process somewhere. The HoloPort is able to do holo-auth (registration email generated, basic ZT entry registered with nodeID and no pubKey/email). User can be taken to hpstatus page.

Holo agent appears to be broken - HoloPort does NOT get registered in `holo-router-registry-agent-id-to-ipv4` Cloudflare KV store.

Unclear how certificate issuance is connected to this step.

We will close this PR for now and revisit as part of pre-launch security / hardening.

PR #589, opened 2020-10-15
Fixed hosted hApp configurations based on DL-0009

mjbrisebois:
The 2 Holofuel versions used in this example are the ones from the example in DL0009.

### DIFF
--- a/modules/services/holochain-conductor.nix
+++ b/modules/services/holochain-conductor.nix
@@ -13,6 +13,9 @@ in
     config = mkOption {
       type = types.attrs;
     };
+    available-happs = mkOption {
+      type = types.listOf types.attrs;
+    };
 
     package = mkOption {
       default = pkgs.holochain-rust;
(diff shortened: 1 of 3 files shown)


zo-el (changes requested):
This is an implementation on `holochain-redux`, but due to the changes in `holochain-rsm` we need a different approach to DL-0009.  Something like the approach @peeech has been thinking/working on for self-hosted happs can be referenced.

Also, we should be branching off of  `2020-10-01-switch-to-RSM` that PJ is using as a base for the rsm release and not `develop`.

mjbrisebois:
> This is an implementation on `holochain-redux`, but due to the changes in `holochain-rsm` we need a different approach to DL-0009. Something like the approach @peeech has been thinking/working on for self-hosted happs can be referenced.
> 
> Also, we should be branching off of `2020-10-01-switch-to-RSM` that PJ is using as a base for the rsm release and not `develop`.

The only change for RSM is how we communicate the resulting app lists to Conductor.  With redux it was through TOML, with RSM it's an API.  This minor change does not invalidate DL-0009.

cc @zo-el

zo-el:
> > This is an implementation on `holochain-redux`, but due to the changes in `holochain-rsm` we need a different approach to DL-0009. Something like the approach @peeech has been thinking/working on for self-hosted happs can be referenced.
> > Also, we should be branching off of `2020-10-01-switch-to-RSM` that PJ is using as a base for the rsm release and not `develop`.
> 
> The only change for RSM is how we communicate the resulting app lists to Conductor. With redux it was through TOML, with RSM it's an API. This minor change does not invalidate DL-0009.
> 
> cc @zo-el

Ok, Let's PL this.
cc: @mjbrisebois @SirRobLyon

PR #592, opened 2020-11-02
Bumps holochain to rev becbc8f

peeech:
It also requires:
- bumping rust toolchain to version 1.47.0
- switching to `.yaml` format of holochain config


peeech:
This work will come for free with 20.09 bump

PR #593, opened 2020-11-04
[switch-to-RSM] Bump NixOS to 20.03

Br1ght0ne:
As per https://github.com/Holo-Host/holo-nixpkgs/pull/577#issuecomment-721706302, this is #577 rebased onto `2020-10-01-switch-to-RSM`.

Additional changes:

- both tests were disabled (needs review and separate enabling commit/PR)


Br1ght0ne:
Closed in favor of #595, which incorporates #594.

PR #594, opened 2020-11-04
nixpkgs: bump to 20.09

Br1ght0ne:
- Remove magic-wormhole package and module
- Remove last patch to nixpkgs
- Rename services.postgresql.{extraConfig -> settings}
- Import nixpkgs directly (skip derivation)


Br1ght0ne:
Closed in favor of #595.

PR #596, opened 2020-11-05
host console ui as derivation

peeech:
Instead of using buggy npm2nix function I used yarn2nix function that reads from yarn.lock. How simple is that?

### DIFF
--- a/overlays/holo-nixpkgs/default.nix
+++ b/overlays/holo-nixpkgs/default.nix
@@ -73,7 +73,9 @@ let
   };
 in
 
-{
+{ 
+  yarn2nix = yarn2nix-moretea;
+ 
   inherit (callPackage aorura {})
     aorura-cli
     aorura-emu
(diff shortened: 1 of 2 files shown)


overlays/holo-nixpkgs/default.nix
```diff
@@ -73,7 +73,9 @@ let
   };
 in
 
-{
+{ 
+  yarn2nix = yarn2nix-moretea;
```
robbiecarlton, line 77: why did you have to bring in yarn?

robbiecarlton (approved):

robbiecarlton:
Looks good and compiles for me. One question about yarn but not a blocker.

overlays/holo-nixpkgs/default.nix
peeech: I switched from npm2nix function of nix to yarn2nix. Both of them serve the same purpose of building js packages into nix derivations. Npm2nix is not maintained and would require debugging and patching work. This is an error that you were seeing while trying to create derivation - npm2nix was not able to handle a corner case of babel. Yarn2nix on the other hand became a part of nixOs starting 20.03, so now that we're updating HPOS to 20.09 it is gonna be available out of the box, maintained. @robbiecarlton

PR #595, opened 2020-11-05
[switch-to-RSM] Bump NixOS to 20.09

Br1ght0ne:

PR #599, opened 2020-11-10
treewide: fix hpos-config build

peeech:


alastairong:
hpos-config builds successfully. Other hydra errors will be fixed in #595

PR #602, opened 2020-11-12
hpos holochain api

robbiecarlton:


alastairong:
@robbiecarlton Can we please name this something that disambiguates it from hpos-admin-api, which is also a host console server :)

Maybe - holochain-hpos-API?

overlays/holo-nixpkgs/default.nix
```diff
@@ -323,6 +323,8 @@ in
 
   inherit (callPackage ./self-hosted-happs {}) self-hosted-happs-node;
 
+  inherit (callPackage ./hpos-holochain-api {}) hpos-holochain-api;
```
robbiecarlton, line 326: is there a namespace conflict here with both of these being called `hpos-holochain-api`?

profiles/logical/hpos/default.nix
```diff
@@ -115,6 +117,13 @@ in
           proxyWebsockets = true;
         };
 
+        "/hcs/v1/" = {
```
peeech, line 120: Can you check if `/hcs/v1/` is desired endpoint name here? @robbiecarlton

overlays/holo-nixpkgs/default.nix
peeech: `./hpos-holochain-api` is a string with a value `./hpos-holochain-api` which is a path to the package, `hpos-holochain-api` is a var name in this scope. So no, they won't be conflicting.

profiles/logical/hpos/default.nix
robbiecarlton: hmm.

That stood for "Host Console Server" which it's no longer called.

It's now called "Hpos Holochain API", ie "/hha", which is going to be confusing

calling it "/holochain" also seems confusing

"/holochain-api" also confusing, but that's probably the best without it being super long.

I'm open to suggestions, otherwise let's just go with "/holochain-api/v1"

peeech: in that case I am going with `/holochain-api/v1`

profiles/logical/hpos/default.nix
```diff
@@ -142,7 +151,7 @@ in
     };
 
     virtualHosts.localhost = {
-        locations."/".proxyPass = "http://unix:/run/hpos-admin.sock:/";
+        locations."/".proxyPass = "http://unix:/run/hpos-admin-api/hpos-admin-api.sock:/";
```
robbiecarlton, line 154: Does this need an equivalent line for `http://unix:/run/hpos-holochain-api/hpos-holochain-api.sock:/`?

peeech: No, because it is used only in tests. This line will actually go away with HPOS tests revamp, because it should be in tests profile, not here

overlays/holo-nixpkgs/hpos-holochain-api/src/index.js
```diff
@@ -0,0 +1,49 @@
...
+
+  const presentedHapps = happs.map(happ => ({
+    id: happ.happ_id,
+    name: happ.happ_bundle.name
+  }))
+
+  res.send(presentedHapps)
+})
+
+app.listen(UNIX_SOCKET, () => {
```
peeech, line 47: First we need to remove stale `UNIX_SOCKET` file if it exists. Otherwise it will throw an error:
```
Error: listen EADDRINUSE: address already in use /run/hpos-holochain-api/hpos-holochain-api.sock
```

zippy (approved):
LGTM, note one tiny version bump for npm holochain-conductor-api  module.  This will certainly bump again...

overlays/holo-nixpkgs/hpos-holochain-api/src/index.js
peeech: fixed in 0e69cf2

peeech (approved):
👍

Br1ght0ne (approved):

PR #605, opened 2020-11-13
holochain: bump to rev ddda967

peeech:

PR #607, opened 2020-11-13
[B-03587] switch self-hosted-happs -> hpos-configure-holochain

zippy:
- [x] update service
- [x] update derivation
- [x] update holochain version to match what's needed in hpos-configure-holochain
- [x] update networking config data in holochain service profile


peeech:
All the valuable work from this PR has been cherry-picked into #605

PR #611, opened 2020-11-19
self-hosted-happs -> configure-holochain

peeech:
Fresh start for switching from node version to rust version of the script


overlays/holo-nixpkgs/configure-holochain/default.nix
```diff
@@ -0,0 +1,28 @@
+{ pkgs, rustPlatform  }:
+
+with pkgs;
+
+let
+  inherit (rust.packages.nightly) rustPlatform;
```
Br1ght0ne, line 6: This should be left out.

overlays/holo-nixpkgs/default.nix
```diff
@@ -122,6 +122,10 @@ in
     wormhole-relay = holo.buildProfile "wormhole-relay";
   };
 
+  configure-holochain = callPackage ./configure-holochain {
```
Br1ght0ne, line 125: When using `packageName = callPackage ...` form, the top-level export in `configure-holochain/default.nix` should be `rustPlatform.buildRustPackage`, not `{ configure-holochain = ... }`.

Br1ght0ne (changes requested):

peeech:
Cherry-picked into #657

PR #615, opened 2020-11-20
Remove Holochain-rust endpoints

alastairong:
@zo-el I think you added the service logger endpoints a while back. They're now obsolete right? And in any case, I imagine we would want Host Console UI to call service logger via the other HPOS-Holochain-API that Robbie wrote?

### DIFF
--- a/modules/services/holochain.nix
+++ b/modules/services/holochain.nix
@@ -32,15 +32,17 @@ in
       requires = [ "lair-keystore.service" ];
       wantedBy = [ "multi-user.target" ];
 
+      #environment.RUST_LOG = "debug";
+
       preStart = ''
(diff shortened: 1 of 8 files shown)


zo-el:
> @zo-el I think you added the service logger endpoints a while back. They're now obsolete right? And in any case, I imagine we would want Host Console UI to call service logger via the other HPOS-Holochain-API that Robbie wrote?

Yes, that is obsolete. The new version of servicelogger does not include an endpoint that returns information that the `hpos-holochain-api` would need. This was excluded as a part of the MVP design of servicelogger.

zo-el (approved):
Looks good

alastairong:
Note: Builds correctly on hydra. Failed check is because it is picking up an aborted build on hydra-2003

zippy (approved):

peeech (approved):
🥇

PR #621, opened 2020-11-23
profiles/hpos: bump stateVersion to 20.09

Br1ght0ne:


peeech:
Hydra is not complaining, but I think that we might wanna do a quick QC test on this one.

peeech (approved):
🥇

alastairong:
Tested as working. Please merge in

PR #623, opened 2020-11-23
profiles/hpos: bump stateVersion to 20.09

alastairong:


peeech (approved):
🥇

alastairong:
Deleting as this will be bundled into self-hosted elemental chat release

PR #626, opened 2020-11-25
Deploying host-console UI

alastairong:
This is to take stock of where we are vs. the host console v0.1 release, from QA point of view


alastairong:
Not sure why hydra is saying this is failing. It's built successfully on both hydra instances...

alastairong:
Closing - using #713 instead

PR #633, opened 2020-11-27
Switch nano kernel version to follow default NixOS version

### LINKED ISSUE #628: HoloPort Nano ethernet broken in 20.09
Tested with a 20.09 image booting directly from SD card.

In NixOS 20.09, Nano does not have working ethernet. `ifconfig` shows an eth0 ip address of `169.25.X.X` which means that DHCP server isn't found. pinging 8.8.8.8 does not work.

This is 99% a kernel issue as reverting to kernel version `5.4` from 19.09 fixes networking. 20.09 kernel is `5.8`. 

CC: @Br1ght0ne @peeech

alastairong:
Currently nano is set to follow latest kernel. This was done back when old kernel (v5.1) did not have good support for Allwinner A64. Current LTS 5.4 which x64 HoloPorts follow has good support.

This fixes #628

### DIFF
--- a/overlays/holo-nixpkgs/default.nix
+++ b/overlays/holo-nixpkgs/default.nix
@@ -146,9 +146,7 @@ rec {
     inherit (rust.packages.holochain-rsm) rustPlatform;
   };
 
-  holoport-nano-dtb = callPackage ./holoport-nano-dtb {
-    linux = linux_latest;
-  };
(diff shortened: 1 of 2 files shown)


peeech (approved):
I like this simplification

peeech:
@alastairong feel free to merge if you see this fix working as expected on nano

alastairong:
@peeech I can't merge this myself. Can you do it? Have confirmed it works fine

Issue #635, opened 2020-12-09
Holochain interfaces are not up after holochain restart


peeech:
`Holochain.service` is configured with help of `holochain-configure.service` after each nix rebuild. This altered state of holochain does not survive holochain restart. 

What needs to happen is after each restart of `holochain.service` a configuration script in form of `holochain-configure.service` needs to be activated. This should happen on a systemd level.

peeech:
It seems that there's no simple fix to this issue with help of systemd options.

Expected behavior of `holochain-configure.service` is:
- whenever `holochain.service` restarts (because of failure and `Restart=` service option or because of `nixos-rebuild`) `configure-holochain.service` should run right after

None of the options in `systemd.unit` (like `Required=`, `PartOf=`, `BindsTo`) guarantees that dependent unit will run after parent unit fails and restarts. It only guarantees that dependent unit will run after parent unit is started.

peeech:
This turns out to be a holochain issue where state that's responsible for open interfaces does not survive restart. Closing.

PR #636, opened 2020-12-09
Configure holochain at each restart

peeech:
tries to address #635

### DIFF
--- a/modules/services/self-hosted-happs.nix
+++ b/modules/services/self-hosted-happs.nix
@@ -29,7 +29,7 @@ in
 
     systemd.services.self-hosted-happs = {
       after = [ "network.target" "holochain.service" ];
-      requires = [ "holochain.service" ];
+      partOf = [ "holochain.service" ];
       wantedBy = [ "multi-user.target" ];
 
       preStart = ''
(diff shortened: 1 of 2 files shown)


Br1ght0ne (approved):
LGTM.

modules/services/self-hosted-happs.nix
```diff
@@ -31,6 +31,7 @@ in
       after = [ "network.target" "holochain.service" ];
       bindsTo = [ "holochain.service" ];
       wantedBy = [ "multi-user.target" ];
+      reloadPropagatedFrom = [ "holochain.service" ];
```
Br1ght0ne, line 34: Are there any docs on this option? Can't seem to find anything.

peeech: `man systemd.unit`, but [nixpkgs systemd wrapper](https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/system/boot/systemd.nix) does not support this option. 

Still this won't work the way I'd like it to. I am gonna describe an issue in #635

Br1ght0ne: Can't we just use `serviceConfig` then?

```
serviceConfig.ReloadPropagatedFrom = "holochain.service";
```

peeech:
This is a wrong direction. Closing.

PR #647, opened 2020-12-19
mkHolochainBinary: specificy a default revision

steveeJ:
This allows consumers use the version specified in holo-nixpkgs without
having to manually keep the versions in sync.


peeech (approved):
Great idea!

PR #655, opened 2021-01-08
Install Hosted happs

zo-el:
**Implemented:**
- [x] `/install_hosted_happ` return number of source chains
- [x] Added `hpos-holochain-client` 
- [x] Added tests for `holochain`
- [x] Added tests for `hpos-holochain-api` in `holo-nixpkgs`


overlays/holo-nixpkgs/holochain/default.nix
```diff
@@ -2,10 +2,10 @@
 
 rec {
   mkHolochainBinary = {
-        version ? "2020-12-17"
-      , rev ? "5daac31fe21f5d10c192bbe208f0b11f8ee67d88"
-      , sha256 ? "1zv063ivgwajr0sidaccr4igy67ll87xnadnm3df6glcmqgy1b3h"
-      , cargoSha256 ? "0w0cn23y858q8jqq77nx5ax51hqan3n43rzzb4146k2f0430rjx6"
+        version ? "2021-01-15"
+      , rev ? "ea026def01d1573d5e0edfb7f4e3e9453f88c43e"
```
zo-el, line 6: **Note:** This bump is needed because we are now using the `registerDna` updates

overlays/holo-nixpkgs/hpos-admin-client/hpos-admin-client.py
```diff
@@ -75,12 +75,6 @@ def disable_feature(ctx, profile, feature):
 def get_status(ctx):
     print(request(ctx, 'GET', '/status').json())
 
-
-@cli.command(help='Get info on happs currently hosted')
```
zo-el, line 79: **Note:**  This endpoint was moved to hpos-holochain-api

profiles/logical/hpos/default.nix
```diff
@@ -166,7 +166,7 @@ in
     '';
   };
 
-  services.holochain = {
+  services.holochain = lib.mkDefault {
```
zo-el, line 169: **Note:** These are set to default so that we can override them in the sandbox setup
cc: @peeech

zo-el:
Another thought about the `hpos-holochain-api` work is that we can move it to its own repo so that we can run the express tests on a CI
cc: @zippy @peeech.

tests/hpos-holochain-api/default.nix
```diff
@@ -0,0 +1,106 @@
...
+    machine.succeed("systemctl restart holochain.service")
+    machine.wait_for_unit("holochain.service")
+    machine.wait_for_open_port("42233")
+
+    machine.succeed("systemctl start hpos-holochain-api.service")
+    machine.wait_for_unit("hpos-holochain-api.service")
+    machine.wait_for_file("/run/hpos-holochain-api/hpos-holochain-api.sock")
+
+    # TODO: find out why the test hangs when you wait for configure-holochain
+    # machine.wait_for_unit("configure-holochain.service")
```
Br1ght0ne, line 49: `configure-holochain` is a [oneshot systemd service](https://www.freedesktop.org/software/systemd/man/systemd.service.html#Type=). In its [current configuration](https://github.com/Holo-Host/holo-nixpkgs/blob/2021-01-07/holochain-api/modules/services/configure-holochain.nix#L43-L49), `RemainAfterExit` is not set (to `yes`), meaning that the status of `configure-holochain.service` is never equal to `active`, only `activating` or `deactivating`.
For `machine.wait_for_unit` to work, `RemainAfterExit = "yes"` needs to be set in `services.configure-holochain.serviceConfig`.

tests/hpos-holochain-api/default.nix
```diff
@@ -0,0 +1,106 @@
...
+  name = "hpos-holochain-api";
+
+  machine = {
+    imports = [ (import "${hpos.logical}/sandbox") ];
+
+    documentation.enable = false;
+
+    environment.systemPackages = [
+      hpos-holochain-client
+      hpos-config  # TODO: now inside hpos-config package
```
Br1ght0ne, line 13:
```suggestion
      hpos-config
```

tests/holochain/default.nix
```diff
@@ -0,0 +1,36 @@
...
+
+makeTest {
+  name = "holochain";
+
+  machine = {
+    imports = [ (import "${hpos.logical}/sandbox") ];
+
+    environment.systemPackages = [
+      # holo-cli
+      hpos-config # TODO: now inside hpos-config package
```
Br1ght0ne, line 11:
```suggestion
      hpos-config
```

Br1ght0ne (commented):
Great work @zo-el! I left a few comments.

Remaining things to address:
- how can we run tests that require the sandbox to be disabled only when we're running under sandbox?

tests/hpos-holochain-api/default.nix
```diff
@@ -0,0 +1,106 @@
...
+      jq
+    ];
+
+    services.hpos-holochain-api.enable = true;
+    services.holochain.config.enable = true;
```
Br1ght0ne, line 18: Pretty sure that `enable` is not an option for Holochain's config file. Maybe you meant `services.holochain.enable`, which should be already set to `true` by the profile?

overlays/holo-nixpkgs/hpos-holochain-api/src/index.js
```diff
@@ -1,48 +1,106 @@
...
-app.get('/hosted_happs', async (_, res) => {
-  const appWebsocket = await AppWebsocket.connect(`ws://localhost:${argv.appPort}`)
+    // Steps:
+    // - Call hha to get happ details
+    let happBundleDetails;
+    try {
+      const APP_ID = await getAppIds()
+      happBundleDetails = await callZome(APP_ID.HHA, 'hha', 'get_happ', happId)
+    } catch (e) {
+      res.sendStatus(500)
```
zippy, line 64: Often times there's a way to put a message into the response if there's a 500 error.  Can we do that?  I think it will help us in debugging.

overlays/holo-nixpkgs/hpos-holochain-api/src/index.js
```diff
@@ -1,48 +1,106 @@
...
-    cell_id: cellId,
-    zome_name: 'hha',
-    fn_name: 'get_happs',
-    provenance: agentKey,
-    payload: {}
-  })
+      // check if the hosted_happ is already listOfInstalledHapps
+      if (listOfInstalledHapps.includes(`${happBundleDetails.happ_id}`)) {
+        console.log(`${happBundleDetails.happ_id}:${dnas[i].nick} already listOfInstalledHapps`)
+        res.sendStatus(501);
```
zippy, line 86: ditto re message if possible.

overlays/holo-nixpkgs/hpos-holochain-api/src/index.js
```diff
@@ -1,48 +1,106 @@
...
+        const serviceloggerPref = parsePreferences(preferences, happBundleDetails.provider_pubkey)
+        console.log("Parsed Preferences: ", serviceloggerPref);
 
-  const presentedHapps = happs.map(happ => ({
-    id: happ.happ_id,
-    name: happ.happ_bundle.name
-  }))
-
-  res.send(presentedHapps)
+        await installHostedDna(happBundleDetails.happ_id, dnas, hostPubKey, serviceloggerPref)
```
zippy, line 91: Can we change the function name to `installHostedHapp` to be more accurate?

overlays/holo-nixpkgs/hpos-holochain-api/src/api.js
```diff
@@ -0,0 +1,159 @@
+const { ADMIN_PORT, HAPP_PORT, getAppIds } = require("./const")
+const { AdminWebsocket, AppWebsocket } = require("@holochain/conductor-api")
+const { downloadFile } = require('./utils')
+const msgpack = require('@msgpack/msgpack')
+
+// NOTE: this code assumes a single DNA per hApp.  This will need to be updated when the hApp bundle
+// spec is completed, and the hosted-happ config Yaml file will also need to be likewise updated
+const installHostedDna = async (happId, dna, agentPubKey, serviceloggerPref) => {
```
zippy, line 8: I wonder if it's worth adding a UUID parameter here (that can be null) so we can be setting the UUID for hostedHapps at install time.  This might be useful for us in testing?  Maybe we don't do it now until we specify how it would be added to the source configuration file.  Just a thought.

zippy (approved):

zippy (changes requested):
Looks very good.  Some changes requested in the comments and some questions.

overlays/holo-nixpkgs/hpos-holochain-api/src/api.js
```diff
@@ -0,0 +1,159 @@
...
+
+        console.log(`Starting app interface on port ${HAPP_PORT}`);
+        await adminWebsocket.attachAppInterface({ port: HAPP_PORT });
+    } catch(e) {
+        console.log(`Error: ${e.message}, probably interface already started.`);
+    }
+}
+
+const callZome = async (installed_app_id, zome_name, fn_name, payload ) => {
+  const appWebsocket = await AppWebsocket.connect(`ws://localhost:${HAPP_PORT}`)
```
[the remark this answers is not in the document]
zo-el: Done. Please look at #665.

overlays/holo-nixpkgs/hpos-holochain-api/src/index.js
zo-el: The reason I went this way was that we are not installing UI's from the happ bundle.
But I guess installing Hosted Happs currently only includes installing DNA's, so it is appropriate to call it `installingHostedHapp`

Ill make the changes

overlays/holo-nixpkgs/hpos-holochain-api/src/api.js
zo-el: We choose to use it we would also have to include the UUID in the happ bundle struct the Holo-Hosting-App, because we cannot just add any UUID since every host should be installing with the same UUID

overlays/holo-nixpkgs/hpos-holochain-api/src/index.js
zo-el: Implemented in PR #665

overlays/holo-nixpkgs/hpos-holochain-api/src/index.js
zo-el: Implemented in PR #665

overlays/holo-nixpkgs/configure-holochain/default.nix
```diff
@@ -7,8 +7,8 @@ rustPlatform.buildRustPackage {
   src = fetchFromGitHub {
     owner = "Holo-Host";
     repo = "hpos-configure-holochain";
-    rev = "e52ac50acc31e7e8dcddab0ff0a8cbd90ae3d6b6";
-    sha256 = "10mk138y2hl5xgpl110yvspvrjcs5p341ixrn7avlziw3cp52fnf";
+    rev = "cade6a3c0638a7dac0634298e64ebb5ec6ade70e";
```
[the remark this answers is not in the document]
zo-el: Should I make this change in this PR or should we update it later?

tests/hpos-holochain-api/default.nix
zo-el: Should we add `RemainAfterExit = "yes"`? Because it makes the tests more reliable. 
It's not necessarily crucial right now. We can also keep this in mind if our tests do start failing.

peeech (approved):
Looks great to me now that all the tests are passing.

peeech (approved):
👍

zippy (approved):

PR #658, opened 2021-01-12
Match engine reads published happ data from HHA and posts it to DOMAIN2HAPP KV store

alastairong:


Br1ght0ne:
Service and packaging code LGTM. Didn't run.

alastairong:
I'm going to merge it to the other match engine branch as I have deployed it and tested that it works. Nothing will pass tests since hydra can't build from the private repos.

alastairong:
The match engine code/architecture is large enough that it should be reviewed together as part of a handover, not via a PR

PR #664, opened 2021-01-18
DevTest: EC and configure-holochain updates

zippy:
- bumps holochain to latest version (includes gossip fixes and RegisterDna updates)
- uses hpos-configure-holochain with UUID
- uses elemental chat ui alpha20
- sets uuid for elemental chat so we have a different hash-space


Br1ght0ne (approved):
Diff LGTM, Hydra build is [passing](https://hydra.holo.host/eval/658099).

peeech:
This is a PR against develop. That means that before merging we need to update uuid of elemental-chat dna. Otherwise develop is gonna share the same hash space with develop

peeech (approved):
Great!

PR #668, opened 2021-01-21
tests: fix hpos-admin-api test

Br1ght0ne:
When #655 is merged, this will be rebased and retargeted onto `develop`.


alastairong:
Looks good and passes test. I am curious about why the diff shows that we used to have a bunch of other tests that were removed?

Br1ght0ne:
> we used to have a bunch of other tests that were removed

The subtest made use of `/hosted_happs` endpoint, which is no longer provided by `hpos-admin-api`.

@alastairong

peeech (approved):
This definitely fixes `hpos-admin-api` test. I have a few issues with the way tests are run currently though which I am going to address in a separate PR.

PR #675, opened 2021-01-26
Holochain restart timer hot fix

alastairong:
Adds a 2 hour restart timer to holochain service, as a hacky stop gap to networking performance degradation.

Note: we will need to reduce this frequency with hosted happs due poor UX.


Br1ght0ne (approved):

peeech (approved):
🥇

PR #677, opened 2021-01-26
Holochain restart timer hot fix to DEVELOP

peeech:
This is a content of PR #675 against develop. Once #675 is confirmed to do what it is supposed to this PR should be merged into develop.


alastairong (approved):

alastairong:
@peeech This has been merged into `master` so I think we should probably merge it to `develop` too

alastairong:
Closing as we want to move away from holochain restarts due to spikey proxy load.

PR #678, opened 2021-01-26
Fix hpos tests

peeech:
This PR is addressing following issues with HPOS tests:
- [x] fix order of `wait_for_port` (a source of randomly flaky tests)
- [x] disable build of virtualbox VM (adds ~6 min to build time and is not in use anymore)
- [x] move test machine settings to separate profile (allows to start testing VM)
- [x] remove holochain test (covered in other tests)
- [x] move all tests to one file (tests do not affect each other, if run in one VM it saves time)
- [x] add possibility for running `hpos-shell` with path to machine profile

Comments welcome.


Br1ght0ne (approved):
LGTM, great work @peeech!

PR #680, opened 2021-01-27
README: add a section on running and writing tests

Br1ght0ne:
If #678 gets implemented, this documentation needs to be updated too.


zo-el (approved):
The content looks good. I like the use of the `<details>` tab too.

peeech (approved):

PR #681, opened 2021-01-28
profiles: add test profile

Br1ght0ne:


tests/hpos-admin-api/default.nix
```diff
@@ -1,47 +1,32 @@
...
-    documentation.enable = false;
-
-    environment.systemPackages = [
-      hpos-admin-client
-      hpos-config
-      jq
-    ];
+    environment.systemPackages = [ hpos-admin-client ];
 
     services.hpos-admin-api.enable = true;
```
peeech, line 11: I am thinking that this line needs to go to test profile as well. The purpose of the test profile is to be able to start VM that is exactly the same as test environment. So all the `service.<>.enable = true;` and `services.nginx.virtualHosts.localhost.locations.<>.proxyPass` should make it to the test profile as well.

peeech (changes requested):

zo-el (commented):
Looks good, I agree with PJ's suggestions. Also, we were going to merge the two tests into one test right? will that be done in a different PR?

Br1ght0ne:
@zo-el

> we were going to merge the two tests into one test

If you mean removing `holochain` test, it's part of #678.

zo-el:
> If you mean removing `holochain` test, it's part of #678.

I mean we were going to merge the `hpos-admin-api` and `hpos-holochain-api` tests into one test file right? so that hydra runs the tests faster?

tests/hpos-admin-api/default.nix
Br1ght0ne: Should be resolved in 47a47f4.

peeech (approved):
Great work!

PR #683, opened 2021-01-29
Fix holoport plus installer

alastairong:
holoport-plus OS is installed on /dev/sdb not /dev/sda


alastairong:
Closing because I borked this PR up by not basing on develop

PR #684, opened 2021-01-29
overlays/hpos-install: change holoport+ disk target from sda to sdb

alastairong:
Manufactured HP Pluses have OS installed on sdb

### DIFF
--- a/overlays/holo-nixpkgs/hpos-install/default.nix
+++ b/overlays/holo-nixpkgs/hpos-install/default.nix
@@ -66,13 +66,13 @@ let
       profile = "<holo-nixpkgs/profiles/targets/holoport-plus>";
 
       prePhase = ''
-        parted /dev/sda --align optimal --script \
+        parted /dev/sdb --align optimal --script \
           mklabel msdos \
           mkpart primary 0% 100% \
           set 1 boot on
 
-        mkfs.ext4 -F /dev/sda1
-        mount /dev/sda1 /mnt
+        mkfs.ext4 -F /dev/sdb1
+        mount /dev/sdb1 /mnt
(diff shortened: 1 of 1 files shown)


alastairong:
Awaiting @marycamacho testing this image on her machine.

Br1ght0ne (approved):
Diff LGTM. Waiting for @marycamacho to test.

overlays/holo-nixpkgs/hpos-install/default.nix
```diff
@@ -66,13 +66,13 @@ let
       profile = "<holo-nixpkgs/profiles/targets/holoport-plus>";
 
       prePhase = ''
-        parted /dev/sda --align optimal --script \
+        parted /dev/sdb --align optimal --script \
```
peeech, line 69: I am not sure what is the logic behind this move. `sda` is a 120GB SSD disk, `sdb` is a 2TB magnetic disk. So IMHO opinion the only way this would make sense was if the system was on `sda` and `/var` was mounted on `sdb`.

alastairong: Well in theory we could change this installer to wipe /sdb and then install the new OS on /sda, BUT then we would have 2 different populations of Holoport Pluses, because factory holoport pluses unfortunately have their OS on /sdb.

peeech: ah ok that would create some serious mess

peeech (approved):

alastairong:
Mary successfully used this on her holoport. 

@peeech @Br1ght0ne let's get this merged in soon

PR #686, opened 2021-01-29
hpos-admin-api: add api usage documentation

alastairong:
The API endpoint documentation belongs here, not in a separate rfcs repo!

Also updated it to reflect the actual endpoints


peeech (approved):
Great. It just reminds me that `/reset` still sends shivers down my spine. It's so easy to wipe entire HP...

alastairong:
Yea. Maybe we should disable it at some point

PR #690, opened 2021-02-03
Pre-release to master

alastairong:
This release is to get 20.09 and updated hpos services to `master` in preparation for elemental chat release.

It is `staging` + 1 commit to disable `/app` endpoint and elemental-chat installation.

/auth endpoint is also disabled to prevent host console login.

Latest develop with tests is not included as it is not part of staging plus the tests are mostly for hosting endpoints


peeech (approved):
approved

zippy (approved):

PR #693, opened 2021-02-04
Envoy integration

alastairong:
Deploy holo-envoy to hpos


overlays/holo-nixpkgs/holo-envoy/default.nix
```diff
@@ -7,15 +7,16 @@
 , ps
 , python
 , fetchgit
+, mkYarnPackage
 }:
 
-stdenv.mkDerivation rec {
+mkYarnPackage rec {
```
zo-el, line 13: I see we have a build script here as well as in the envoy repo. Do we actually need the scripts to be in both places?

zo-el (changes requested):
@alastairong In the envoy repo, we already have a build script in nix. Do we need it in two locations?
This is the work I am referencing: https://github.com/Holo-Host/holo-envoy/pull/53

overlays/holo-nixpkgs/holo-envoy/default.nix
alastairong: I think in general we want the script in `holo-nixpkgs` rather than in `holo-envoy` so that it uses the same nixpkgs etc. as hpos. I just updated in `holo-envoy` because I think we wanted it there too for some other reason e.g. development

alastairong:
Closing - using PR 694

PR #694, opened 2021-02-04
Hha and tools

alastairong:
Adding HHA and SL
Also adding hpos-holochain-client to CLI for manual hosted-happ installation (until we do auto-install)


profiles/logical/hpos/default.nix
```diff
@@ -41,7 +41,7 @@ in
   # REVIEW: `true` breaks gtk+ builds (cairo dependency)
   environment.noXlibs = false;
 
-  environment.systemPackages = [ hc-state hpos-reset hpos-admin-client hpos-update-cli git ];
+  environment.systemPackages = [ hc-state hpos-reset hpos-admin-client hpos-update-cli git hpos-holochain-client ];
```
zo-el, line 44: We do not need the `hpos-holochain-client`. It is just meant for the test environment.

profiles/logical/hpos/default.nix
```diff
@@ -206,7 +207,28 @@ in
...
     working-directory = configureHolochainWorkingDir;
     install-list = {
-      core_happs = [];
+      core_happs = [
+        {
+          app_id = "core-happs";
+          uuid = "0001";
+          version = "alpha9";
+          /* dna_url = "https://holo-host.github.io/holo-hosting-app-rsm/releases/downloads/v0.0.1-alpha6/holo-hosting-app.dna.gz"; */
+          dna_path = builtins.fetchurl {
```
zo-el, line 216: `dna_path` is also meant to be used only for the dev environment. For the live `hpos` you need to pass dna_url instead.

Just for context, the need for the dna_path was that in the test environment we cannot download a file since the hpos-tests are run in the sandbox.
All we have to do here is comment in the `dna_url` variable and just pass the URL

zo-el (changes requested):
I have left a few comments

profiles/logical/hpos/default.nix
alastairong: well right now we need it to actually trigger installation of hosted happs right? Since `configure-holochain` doesn't do that.

Also, in my experience it's been useful to expose the api clients to the CLI right now for general testing purposes - the `hpos-admin-client` is as well. We can remove in future when we're truly live.

profiles/logical/hpos/default.nix
alastairong: Cool. Will switch that

profiles/logical/hpos/default.nix
Br1ght0ne, line 44:
```suggestion
  environment.systemPackages = [ git hc-state hpos-admin-client hpos-holochain-client hpos-reset hpos-update-cli ];
```

Br1ght0ne (approved):

alastairong:
Due to hydra being a bitch today, I'm merging envoy-integration into this story for testing

modules/services/holochain.nix
```diff
@@ -28,7 +28,7 @@ in
     environment.systemPackages = [ cfg.package ];
 
     systemd.services.holochain = {
-      after = [ "network.target" "lair-keystore.service" ];
+      after = [ "network.target" "lair-keystore.service" "holo-envoy.service" ];
       requires = [ "lair-keystore.service" ];
```
zo-el, line 32:
```suggestion
      requires = [ "lair-keystore.service" "holo-envoy.service" ];
```

profiles/logical/hpos/default.nix
```diff
@@ -218,7 +219,28 @@ in
...
     working-directory = configureHolochainWorkingDir;
     install-list = {
-      core_happs = [];
+      core_happs = [
+        {
+          app_id = "core-happs";
+          uuid = "0001";
+          version = "alpha9";
+          dna_url = "https://holo-host.github.io/holo-hosting-app-rsm/releases/downloads/v0.0.1-alpha9/holo-hosting-app.dna.gz";
+          # dna_path = builtins.fetchurl {
```
zo-el, line 228: Should we just clean this up? 
We do not need this commented code here

overlays/holo-nixpkgs/holo-envoy/default.nix
```diff
@@ -7,15 +7,16 @@
 , ps
 , python
 , fetchgit
+, mkYarnPackage
 }:
 
-stdenv.mkDerivation rec {
+mkYarnPackage rec {
```
zo-el, line 13: I still think we should have these build derivations in only one place. (Preferably in the holo-envoy repo, since if the holo-envoy code build process needs to be updated we can update it in the same repo.)

Mostly if we have it in two places we need to maintain two spaces with the same code, unless we see an advantage for it

alastairong: @peeech Can you guys discuss and let me know what you'd prefer? I don't mind either. We moved it from holo-envoy to holo-nixpkgs repo for some reason. I think it was to ensure it used the correct holochain.

alastairong:
@peeech @zo-el We are also going to need to update our hpos tests to be reflective of having envoy and lair shim. I was having trouble doing this locally because my tests seem to be flaky...

Will deal with it next week I guess

overlays/holo-nixpkgs/holo-envoy/default.nix
peeech: Yarn it is

peeech (approved):
This looks good

peeech (approved):

marycamacho:
Using Admin to merge at PJs request.

PR #695, opened 2021-02-04
holochain: add version groups and meta packages

steveeJ:
The motivation here is to make it convenient to follow a specific branch
for all packages related to holochain, or specify a custom revision for
holochain.

### DIFF
--- a/overlays/holo-nixpkgs/holochain/default.nix
+++ b/overlays/holo-nixpkgs/holochain/default.nix
@@ -2,8 +2,7 @@
 
 let
   mkHolochainBinary = {
-      version
-      , rev
+      rev
       , sha256
       , cargoSha256
       , crate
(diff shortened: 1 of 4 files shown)


overlays/holo-nixpkgs/holochain/default.nix
```diff
@@ -48,9 +47,73 @@ rec {
     "sha256"
     "cargoSha256"
     "crate"
+    "bins"
```
freesig, line 50: FIXME: Maybe Remove

freesig (approved):
lgtm

overlays/holo-nixpkgs/holochain/default.nix
steveeJ: turns out we do need to remove here as well

Br1ght0ne (approved):

PR #698, opened 2021-02-09
Master release candidate

zippy:
- reverts ec disabling change on master
- merges in staging
- bumps UUID to 0002 for clean hash-space


profiles/logical/hpos/default.nix
```diff
@@ -191,7 +191,7 @@ in
           };
           proxy_config = {
             type = "remote_proxy_client";
-            proxy_url = "kitsune-proxy://nFCWLsuRC0X31UMv8cJxioL-lBRFQ74UQAsb8qL4XyM/kitsune-quic/h/proxy.holochain.org/p/5775/--";
+            proxy_url = "kitsune-proxy://nFCWLsuRC0X31UMv8cJxioL-lBRFQ74UQAsb8qL4XyM/kitsune-quic/h/165.22.32.11/p/5775/--";
```
alastairong, line 194: Which proxy are we using officially? The DO one? Why are we hardcoding the IP?

alastairong (dismissed):
Looks good to me. Host Console login still disabled. Elemental chat should be re-enabled. One question posted.

peeech (changes requested):
Staging does not represent our develop direction. This discrepancy has to be resolved before we try to merge staging to master

profiles/logical/hpos/default.nix
zippy: DO. Encoded this way because the proxy server can't match on domain name.

tests/default.nix
```diff
@@ -12,6 +12,7 @@ in
 
 {
   # TODO: fix/port/review these
-  # hpos-admin-api = callPackage ./hpos-admin-api {};
-  # holochain-conductor = callPackage ./holochain-conductor {};
+  # holochain = callPackage ./holochain {};
```
zo-el, line 15: If we want to remove the `holochain` test we should remove the test file as well

tests/default.nix
```diff
@@ -12,6 +12,7 @@ in
 
 {
   # TODO: fix/port/review these
```
zo-el, line 14: We can remove this comment

zo-el (dismissed):
LGTM. Just left a few small comments.

peeech (approved):
Now it looks good

Br1ght0ne (approved):

tests/default.nix
peeech: that's fine for now because NixOs is lazy and won't try to evaluate

peeech (approved):
👍

alastairong (approved):

PR #710, opened 2021-02-18
Give apis group write permissions on USB

peeech:


peeech:
Closing as writing to USB drive that contains sensitive info is considered insecure

PR #728, opened 2021-03-02
Update to holochain with MapFull fix and set bigger value in env var

zippy:


alastairong (dismissed):
Looks fine to me

peeech (approved):
:+1:

peeech (approved):
again 👍

PR #732, opened 2021-03-05
overlays/holo-nixpkgs/configure-holochain: bumped to e4fc3

zo-el:


alastairong:
Approving here to review as part of #730

PR #744, opened 2021-03-15
Update HPOS to allow an option for joining the scalability testing network

alastairong:


alastairong:
This doesn't work. Closing

PR #747, opened 2021-03-16
Add a flashing orange state to highlight holochain connectivity issues

alastairong:
If LED manager cannot make a proxy-cli call to the proxy server, it will start flashing orange

I tested this by giving it a fake proxy address and it seems to work fine.

### DIFF
--- a/overlays/holo-nixpkgs/hpos-led-manager/src/main.rs
+++ b/overlays/holo-nixpkgs/hpos-led-manager/src/main.rs
@@ -58,7 +58,7 @@ fn main() -> Fallible<()> {
             .args(&["--", &args.flag_kitsune])
             .output()
         {   
-            Ok(output) =>{ 
+            Ok(output) => { 
                 let output_string = String::from_utf8(output.stdout)?;
                 output_string.contains("tokio_task_count")
             },
--- a/overlays/holo-nixpkgs/hpos-led-manager/src/main.rs
+++ b/overlays/holo-nixpkgs/hpos-led-manager/src/main.rs
@@ -44,6 +44,8 @@ fn main() -> Fallible<()> {
     let state_path = args.flag_state;
     let state_temp_path = state_path.with_extension("tmp");
 
+    let mut counter: u8 = 0;
+
     loop {
         let router_gateway_addrs = "router-gateway.holo.host:80".to_socket_addrs();
         let conn_zerotier = match router_gateway_addrs {
@@ -54,16 +56,24 @@ fn main() -> Fallible<()> {
             Err(_) => false,
         };
         
-        let conn_kitsune_proxy = match Command::new("proxy-cli")
-            .args(&["--", &args.flag_kitsune])
-            .output()
-        {   
-            Ok(output) => { 
-                let output_string = String::from_utf8(output.stdout)?;
-                output_string.contains("tokio_task_count")
-            },
-            Err(_) => false,
+        let mut conn_kitsune_proxy = true;
+
+        if counter % 30 == 0 {
+            conn_kitsune_proxy = match Command::new("proxy-cli")
+                .args(&["--", &args.flag_kitsune])
+                .output()
+            {   
+                Ok(output) => { 
+                    let output_string = String::from_utf8(output.stdout)?;
+                    output_string.contains("tokio_task_count")
+                },
+                Err(_) => false,
+            };
+
+            counter = 0;
         };
+        
+        counter = counter + 1;
 
         let hpos_config_found = Path::new("/run/hpos-init/hpos-config.json").exists();
 
--- a/overlays/holo-nixpkgs/hpos-led-manager/src/main.rs
+++ b/overlays/holo-nixpkgs/hpos-led-manager/src/main.rs
@@ -14,7 +14,7 @@ use std::process::Command;
 const POLLING_INTERVAL: u64 = 1;
 
 const USAGE: &'static str = "
-Usage: hpos-led-manager --device <path> --state <path> --kitsune_address <url>
+Usage: hpos-led-manager --device <path> --state <path> --kitsune <url>
        hpos-led-manager --help
 
 Manages AORURA LED.
@@ -22,14 +22,14 @@ Manages AORURA LED.
 Options:
   --device <path>  Path to AORURA device
   --state <path>   Path to state JSON file
-  --kitsune_address Kitsune proxy URL to check
+  --kitsune <url>  Kitsune proxy url to check
 ";
 
 #[derive(Debug, Deserialize)]
 struct Args {
     flag_device: PathBuf,
     flag_state: PathBuf,
-    kitsune_address: String
+    flag_kitsune: String
 }
 
 fn main() -> Fallible<()> {
@@ -55,7 +55,7 @@ fn main() -> Fallible<()> {
         };
         
         let conn_kitsune_proxy = match Command::new("proxy-cli")
-            .args(&["--", &args.kitsune_address])
+            .args(&["--", &args.flag_kitsune])
             .output()
         {   
             Ok(output) =>{ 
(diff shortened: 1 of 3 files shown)


overlays/holo-nixpkgs/hpos-led-manager/src/main.rs
```diff
@@ -43,19 +46,34 @@ fn main() -> Fallible<()> {
...
+        let conn_kitsune_proxy = match Command::new("proxy-cli")
+            .args(&["--", &args.flag_kitsune])
+            .output()
+        {   
+            Ok(output) =>{ 
+                let output_string = String::from_utf8(output.stdout)?;
+                match output_string {
+                    _ if output_string.contains("tokio_task_count") => true,
+                    _ => false,
+                }
```
timotree3, line 66:
```suggestion
                output_string.contains("tokio_task_count")
```

overlays/holo-nixpkgs/hpos-led-manager/src/main.rs
```diff
@@ -43,19 +46,34 @@ fn main() -> Fallible<()> {
...
                 None => false,
             },
             Err(_) => false,
         };
+        
+        let conn_kitsune_proxy = match Command::new("proxy-cli")
+            .args(&["--", &args.flag_kitsune])
+            .output()
+        {   
+            Ok(output) =>{
```
timotree3, line 61:
```suggestion
            Ok(output) => { 
```

timotree3 (approved):
I approve of the rust changes with a couple of style suggestions. (can't speak for the nix)

Br1ght0ne (approved):
Nix looks good, and thanks @timotree3 for the review! LGTM.

profiles/logical/hpos/default.nix
```diff
@@ -41,7 +43,7 @@ in
   # REVIEW: `true` breaks gtk+ builds (cairo dependency)
   environment.noXlibs = false;
 
-  environment.systemPackages = [ git hc-state hpos-admin-client hpos-holochain-client hpos-reset hpos-update-cli ];
+  environment.systemPackages = [ git hc-state hpos-admin-client hpos-holochain-client hpos-reset hpos-update-cli holochainAllBinariesWithDeps.hpos.kitsune-p2p-proxy ];
```
peeech, line 46: My question here is why do we need to add `holochainAllBinariesWithDeps.hpos.kitsune-p2p-proxy ` to system packages on HPOS?

overlays/holo-nixpkgs/hpos-led-manager/src/main.rs
```diff
@@ -43,19 +46,31 @@ fn main() -> Fallible<()> {
...
-        let online = match router_gateway_addrs {
+        let conn_zerotier = match router_gateway_addrs {
             Ok(mut addrs) => match addrs.next() {
                 Some(addr) => TcpStream::connect_timeout(&addr, Duration::new(1, 0)).is_ok(),
                 None => false,
             },
             Err(_) => false,
         };
+        
+        let conn_kitsune_proxy = match Command::new("proxy-cli")
```
peeech, line 57: My only concern here is that this code is executed in a loop every 1s. Assuming 2000 holoports out there it adds 2k/s requests on proxy. Same actually applies to zerotier connection check, where `router-gateway.holo.host` receives 2k/s connections just for a status check. I wonder if we should be worried about that?

profiles/logical/hpos/default.nix
alastairong: My understanding was that the Command::new function in the LED manager code basically runs a shell command and therefore the proxy-cli command needed to be available to the shell. I didn't dig into it deeply so I may have misunderstood

overlays/holo-nixpkgs/hpos-led-manager/src/main.rs
alastairong: hmmm maybe we reduce the frequency...let me ask

profiles/logical/hpos/default.nix
peeech: oh, ok. What you are talking about is achieved in [this line](https://github.com/Holo-Host/holo-nixpkgs/blob/LED-update/modules/services/hpos-led-manager.nix#L36) where you add `proxy-cli` to the PATH of `hpos-led-manager.service` that wants to execute `proxy-cli`.

What the line under discussion does it [adds a package](https://search.nixos.org/options?channel=20.09&from=0&size=50&sort=relevance&query=environment.systemPackages) (and its executable binaries like `proxy-cli`) to global environment. As a result all the packages listed here are available in ie bash - so that you can call them when you ssh into your machine.

I think it's unnecessary here.

alastairong: Well Zippy was saying it's a helpful tool for diagnostics anyway, so perhaps we should include it anyway?

peeech: #753 is my take on that. I think that if we want to use HPOS as a dev machine (which I do all the time) then having holochain bins available for ssh-ed user is a must. In #753 I am adding all holochian bins. `dna-util` will be replaced with `hc` ASAP.

overlays/holo-nixpkgs/hpos-led-manager/src/main.rs
alastairong: I will switch this to once every 30 seconds. I won't change the ZT frequency.

peeech: This will change frequency of both checks. The other duration `Duration::new(1, 0)` is a timeout for TCP call.

overlays/holo-nixpkgs/hpos-led-manager/src/main.rs
```diff
@@ -41,21 +44,42 @@ fn main() -> Fallible<()> {
...
+                    output_string.contains("tokio_task_count")
+                },
+                Err(_) => false,
+            };
+
+            counter = 0;
+        };
+        
+        counter = counter + 1;
+
```
alastairong, line 76: @timotree3 Can I get your review on the Rust syntax here? I want is to check proxy connectivity every 30 seconds and everything else every 1 second. In between those 30 seconds, the proxy check should just continue to show the result of the last check.

overlays/holo-nixpkgs/hpos-led-manager/src/main.rs
```diff
@@ -41,21 +44,42 @@ fn main() -> Fallible<()> {
...
+                    let output_string = String::from_utf8(output.stdout)?;
+                    output_string.contains("tokio_task_count")
+                },
+                Err(_) => false,
+            };
+
+            counter = 0;
+        };
+        
+        counter = counter + 1;
```
timotree3, line 75:
```suggestion
        if counter == 0 {
            conn_kitsune_proxy = match Command::new("proxy-cli")
                .args(&["--", &args.flag_kitsune])
                .output()
            {   
                Ok(output) => { 
                    let output_string = String::from_utf8(output.stdout)?;
                    output_string.contains("tokio_task_count")
                },
                Err(_) => false,
            };
        };
        
        counter = (counter + 1) % 30;
```

this is how I would do it but your way is fine too

overlays/holo-nixpkgs/hpos-led-manager/src/main.rs
```diff
@@ -41,21 +44,40 @@ fn main() -> Fallible<()> {
...
+            {   
+                Ok(output) => { 
+                    let output_string = String::from_utf8(output.stdout)?;
+                    output_string.contains("tokio_task_count")
+                },
+                Err(_) => false,
+            };
+        };
+        
+        counter = (counter + 1) % 30;
```
zippy, line 73: This looks like it won't work to me.  This will allways evaluate to 1.

zippy (changes requested):

overlays/holo-nixpkgs/hpos-led-manager/src/main.rs
alastairong: Did I get the syntax wrong for incrementing?

zippy: my bad.

zippy (approved):

PR #750, opened 2021-03-17
Use catchup version configure-holochain

zippy:


alastairong:
@zippy Why is this a separate PR rather than an addition to Joel's PR 749, which seems to be a general "holochain bump" PR?

zippy:
Closing if favor of #749

PR #751, opened 2021-03-20
Update to holochain with new proxy

zippy:
Turns out this doesn't really affect production yet.


zippy:
Actually don't want the commit with "new proxy" because that's not actually new proxy.  Cherry picket the tryorama commit into #750

PR #752, opened 2021-03-22
bump holochian and add libiconv dep on darwin

steveeJ:
Extracted from https://github.com/Holo-Host/holo-nixpkgs/pull/749


Br1ght0ne (approved):
Crocodile Hunter Yes

PR #756, opened 2021-03-25
overlays/holo-nixpkgs/hpos-holochain-api: Catch up to new holochain

timotree3:
Changes:
- Update to use `@holochain/conductor-api@0.1.0`
- Update to use `bundle_url` provided by HHA for installing hosted happs, rather than using each `dna_url` individually.
- Update to use `hc sandbox` instead of `holochain-run-dna`
- Update the versions of servicelogger and HHA that get downloaded for testing
- Fix pre-existing apparent bug where we don't reliably wait for input to come through when handling an HTTP request (Hasn't seemed to show up in practice though)

I did not end up changing the servicelogger cloning mechanism. I tried using the newly introduced `createCloneCell` but it only allows you to clone cells such that they end up inside of an already active app. In the future it might make sense to move all of the servicelogger instances into one app, and use `slotId`s to identify them and `createCloneCell` to clone them.

### DIFF
--- a/overlays/holo-nixpkgs/hpos-holochain-api/src/const.js
+++ b/overlays/holo-nixpkgs/hpos-holochain-api/src/const.js
@@ -24,7 +24,7 @@ const getAppIds = async () => {
   try {
     const config = yaml.load(fs.readFileSync(CONFIGURE_HC, 'utf8'))
     const getId = (name) => {
-      const bundle_url = config.core_happs.find(h => h.bundle_url.includes(name))
+      const { bundle_url } = config.core_happs.find(h => h.bundle_url.includes(name))
       return new URL(bundle_url).pathname
       .replace('.happ', '')
       .replace('.', ':')
--- a/overlays/holo-nixpkgs/hpos-holochain-api/src/const.js
+++ b/overlays/holo-nixpkgs/hpos-holochain-api/src/const.js
@@ -31,7 +31,7 @@ const getAppIds = async () => {
         .replace('.', ':')
     }
     return {
-      HHA: getId('holo-hosting-app'),
+      HHA: getId('core-app'),
       SL: getId('servicelogger')
     }
   } catch (e) {
--- a/overlays/holo-nixpkgs/hpos-holochain-api/src/const.js
+++ b/overlays/holo-nixpkgs/hpos-holochain-api/src/const.js
@@ -7,7 +7,7 @@ const ADMIN_PORT = 4444
 
 const HAPP_PORT = 42233
 
-const CONFIGURE_HC = process.env.NODE_ENV === 'test' ? './tests/config.yml' : '/var/lib/configure-holochain/config.yaml'
+const CONFIGURE_HC = process.env.NODE_ENV === 'test' ? './tests/config.yaml' : '/var/lib/configure-holochain/config.yaml'
 const READ_ONLY_PUBKEY = '/var/lib/configure-holochain/agent_key.pub'
 
 const getReadOnlyPubKey = async () => {
@@ -22,7 +22,7 @@ const getReadOnlyPubKey = async () => {
 
 const getAppIds = async () => {
   try {
-    const config = await yaml.load(fs.readFileSync(CONFIGURE_HC, 'utf8'))
+    const config = yaml.load(fs.readFileSync(CONFIGURE_HC, 'utf8'))
     const getId = (id) => {
       const app = config.core_happs.find(h => h.app_id === id)
       if (app.uuid === undefined) {
--- a/overlays/holo-nixpkgs/hpos-holochain-api/src/const.js
+++ b/overlays/holo-nixpkgs/hpos-holochain-api/src/const.js
@@ -25,9 +25,10 @@ const getAppIds = async () => {
     const config = yaml.load(fs.readFileSync(CONFIGURE_HC, 'utf8'))
     const getId = (name) => {
       const { bundle_url } = config.core_happs.find(h => h.bundle_url.includes(name))
-      return new URL(bundle_url).pathname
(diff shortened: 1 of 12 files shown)


zo-el (changes requested):
Just left a few comments and questions

overlays/holo-nixpkgs/hpos-holochain-api/tests/index.test.js
```diff
@@ -19,12 +19,13 @@ function delay(t, val) {
 test('holochain-api endpoint ', async () => {
   const listOfHappsResponse = await request(app).get('/hosted_happs').send(usageTimeInterval)
   expect(listOfHappsResponse.status).toBe(200)
-  expect(listOfHappsResponse.body[0].name).toBe(HAPP_NAME)
-  expect(listOfHappsResponse.body[0].enabled).toBe(false)
-  expect(listOfHappsResponse.body[0].error.message).toBeTruthy()
-  expect(listOfHappsResponse.body[0].error.source).toBeTruthy()
-  expect(listOfHappsResponse.body[0].sourceChains).toBeFalsy()
-  expect(listOfHappsResponse.body[0].usage).toBeFalsy()
+  const listOfHapps = JSON.parse(listOfHappsResponse.text)
```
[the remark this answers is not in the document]
timotree3: Tests were failing because `body` was `undefined` even though `text` contained a JSON string. It's as though there was some feature of the http library that automatically parsed the response text as json and now it's disabled.

overlays/holo-nixpkgs/hpos-holochain-api/src/const.js
```diff
@@ -22,25 +22,16 @@ const getReadOnlyPubKey = async () => {
...
-        HHA: config[0].app_name,
-        SL: config[1].app_name
-      }
-    } else {
-      return {
-        HHA: getId('core-happs'),
-        SL: getId('servicelogger')
-      }
+    return {
+      HHA: getId('holo-hosting-app'),
```
alastairong, line 33: @zo-el @timotree3 Can you align on the correct name here? I think the new bundle is called `core-app`

timotree3: Not sure. @zo-el?

zo-el: Yes it's called core-app.

timotree3:
```suggestion
      HHA: getId('core-app'),
```

zo-el (approved):
Looks good

peeech (approved):
Looks good

overlays/holo-nixpkgs/hpos-holochain-api/src/index.js
```diff
@@ -58,11 +58,13 @@ const getPresentedHapps = async usageTimeInterval => {
...
 
 app.get('/hosted_happs', async (req, res) => {
-  let usageTimeInterval
-  await req.on('data', (body) => {
-    usageTimeInterval = JSON.parse(body.toString())
-    if (!isusageTimeInterval(usageTimeInterval)) return res.status(501).send('error from /hosted_happs: param provided is not an object')
-  })
+  const usageTimeInterval = await Promise.race([
+    new Promise(resolve => req.on('data', (body) => {
+      resolve(JSON.parse(body.toString()))
```
Connoropolous, line 63: @timotree3 
JSON.parse can easily throw, and it's not in a `catch` block, so I'm thinking someone could crash the server by sending a badly formatted json body

overlays/holo-nixpkgs/hpos-holochain-api/src/index.js
```diff
@@ -73,11 +75,10 @@ app.get('/hosted_happs', async (req, res) => {
 })
 
 app.get('/dashboard', async (req, res) => {
-  let usageTimeInterval
-  await req.on('data', (body) => {
-    usageTimeInterval = JSON.parse(body.toString())
-    if (!isusageTimeInterval(usageTimeInterval)) return res.status(501).send('error from /hosted_happs: param provided is not an object')
-  })
+  const usageTimeInterval = await new Promise(resolve => req.on('data', (body) => {
+    resolve(JSON.parse(body.toString()))
```
Connoropolous, line 79: same thing here

overlays/holo-nixpkgs/hpos-holochain-api/src/index.js
Connoropolous: how come you're using `body` and `req.on` data for a GET request?

Connoropolous: using a body/payload on an HTTP GET request is simply very non standard, I thought it wasn't even allowed, but I looked into it and was wrong, and I'm not one who believes in being super dogmatic anyways, I just was surprised this even worked.
Worth noting that one cannot access this method usefully as a GET request through a browser, because you can't pass the params it calls for as query params or directly in the path

Connoropolous: @timotree3 I wasn't totally sure, so I peeked at the tests and saw I could add one easily: 
here's steps to replicate a crashing request
https://github.com/Holo-Host/holo-nixpkgs/pull/767

Connoropolous: and this wasn't your doing, I can see the JSON.parse was already there, uncaught

Connoropolous: the important thing to remember is that JSON.parse throws an error with a bad input, it doesn't just return `undefined`

PR #758, opened 2021-03-25
Fix tests

alastairong:
- Fixes flakey tests caused by lair needing more time to get ready and api-nginx dependencies. 
- Bumps tests for new holochain
- Bump holo-envoy and enable it for testing
- Add a 2.5 second sleep to envoy start up

I have run this many times without failure :)

### DIFF
--- a/modules/services/hpos-admin-api.nix
+++ b/modules/services/hpos-admin-api.nix
@@ -18,8 +18,8 @@ in
 
   config = mkIf cfg.enable {
     systemd.services.hpos-admin-api = {
-      after = [ "network.target" ];
-      wantedBy = [ "multi-user.target" ];
+      after = [ "network.target" "nginx.service" ];
+      wantedBy = [ "multi-user.target" "nginx.service" ];
 
       serviceConfig = {
         ExecStart = "${cfg.package}/bin/hpos-admin-api";
--- a/modules/services/hpos-admin-api.nix
+++ b/modules/services/hpos-admin-api.nix
@@ -19,7 +19,7 @@ in
   config = mkIf cfg.enable {
     systemd.services.hpos-admin-api = {
       after = [ "network.target" "nginx.service" ];
-      wantedBy = [ "multi-user.target" "nginx.service" ];
+      wantedBy = [ "multi-user.target" ];
 
       serviceConfig = {
         ExecStart = "${cfg.package}/bin/hpos-admin-api";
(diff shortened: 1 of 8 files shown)


modules/services/hpos-admin-api.nix
```diff
@@ -18,8 +18,8 @@ in
 
   config = mkIf cfg.enable {
     systemd.services.hpos-admin-api = {
-      after = [ "network.target" ];
-      wantedBy = [ "multi-user.target" ];
+      after = [ "network.target" "nginx.service" ];
+      wantedBy = [ "multi-user.target" "nginx.service" ];
```
Br1ght0ne, line 22: There's no reason `nginx.service` would "want" this to be started. `wantedBy` makes sense only as a guide to when the service should be started (`multi-user.target` effectively means "system is accessible to multiple users == startup finished").

```suggestion
      wantedBy = [ "multi-user.target" ];
```

modules/services/hpos-holochain-api.nix
```diff
@@ -18,8 +18,8 @@ in
 
   config = mkIf cfg.enable {
     systemd.services.hpos-holochain-api = {
-      after = [ "network.target" ];
-      wantedBy = [ "multi-user.target" ];
+      after = [ "network.target" "nginx.service" ];
+      wantedBy = [ "multi-user.target" "nginx.service" ];
```
Br1ght0ne, line 22:
```suggestion
```

Br1ght0ne (changes requested):
Checks out, except for service configuration nitpicks.

modules/services/hpos-holochain-api.nix
alastairong: LOL I misread that as `required = `. Will address

alastairong:
Ironically, it failed tests. This was because Envoy took 12 seconds to get ready.

I will change from a sleep to a more intelligent wait, since I don't want to add a 15 second forced sleep whenever Envoy restarts

alastairong:
@Br1ght0ne Can i get a re-review please? 

Updated Envoy service to wait until it has started the server, for a max of 1 minute
Also updated hpos-holochain-api to wait for holochain

The envoy handling is probably overkill normally. It takes ages on hydra but not when running locally.

Br1ght0ne (approved):
LGTM, thanks @alastairong!

PR #767, opened 2021-03-30
prevent server crash, in case of bad formatted json

Connoropolous:
I was reading through PRs and came on something that looked like it had a hole
 
The following test failure occurs when I simply added a certain string formatting that's not valid JSON, because a JSON.parse call was not in a try/catch block, and it always should be, especially on the server side
```
  ● holochain-api endpoint

    SyntaxError: Unexpected token b in JSON at position 0
        at JSON.parse (<anonymous>)

      61 |   const usageTimeInterval = await Promise.race([
      62 |     new Promise(resolve => req.on('data', (body) => {
    > 63 |       resolve(JSON.parse(body.toString()))
         |                    ^
      64 |     })),
      65 |     new Promise(resolve => setTimeout(() => resolve(undefined), 100))
      66 |   ])

      at IncomingMessage.<anonymous> (src/index.js:63:20)
```

### DIFF
--- a/overlays/holo-nixpkgs/hpos-holochain-api/src/index.js
+++ b/overlays/holo-nixpkgs/hpos-holochain-api/src/index.js
@@ -57,28 +57,38 @@ const getPresentedHapps = async usageTimeInterval => {
   return presentedHapps
 }
 
-app.get('/hosted_happs', async (req, res) => {
-  const usageTimeInterval = await Promise.race([
+// process the 'body' of the request, and parse it into JSON
+// falling back to returning undefined either if no body is passed
+// or the string fails to parse as valid JSON
+const handleBodyData = (req) => {
+  return Promise.race([
     new Promise(resolve => req.on('data', (body) => {
-      resolve(JSON.parse(body.toString()))
+      try {
+        resolve(JSON.parse(body.toString()))
+      } catch (_) {
+        // parse error
+        resolve(undefined)
+      }
     })), 
     new Promise(resolve => setTimeout(() => resolve(undefined), 100))
   ])
-  if (usageTimeInterval !== undefined && !isusageTimeInterval(usageTimeInterval)) return res.status(501).send('error from /hosted_happs: param provided is not an object')
+}
+
+app.get('/hosted_happs', async (req, res) => {
+  const usageTimeInterval = await handleBodyData(req)
+  if (!isusageTimeInterval(usageTimeInterval)) return res.status(501).send('error from /hosted_happs: param provided is not an object')
 
   try {
     const presentedHapps = await getPresentedHapps(usageTimeInterval)
     res.status(200).send(presentedHapps)
   } catch (e) {
-    return res.status(501).send(`hpos-holochain-api error: ${e}`)
+    return res.status(501).send(`hpos-holochain-api error: ${JSON.stringify(e)}`)
   }
 })
 
 app.get('/dashboard', async (req, res) => {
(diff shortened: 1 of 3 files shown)


Connoropolous:
sure, it's easy enough, I'll fix

Connoropolous:
@timotree3 resolved!

Connoropolous:
I guess CI doesn't actually run these tests? 
Who runs these tests? I mostly ran them locally, but I didn't know which exact holochain commit to be on

alastairong:
Hydra CI runs these tests but the test coverage is just happy path, and is currently disabled for this API

overlays/holo-nixpkgs/hpos-holochain-api/src/index.js
```diff
@@ -57,28 +57,38 @@ const getPresentedHapps = async usageTimeInterval => {
...
+const handleBodyData = (req) => {
+  return Promise.race([
     new Promise(resolve => req.on('data', (body) => {
-      resolve(JSON.parse(body.toString()))
+      try {
+        resolve(JSON.parse(body.toString()))
+      } catch (_) {
+        // parse error
+        resolve(undefined)
+      }
```
timotree3, line 71: Can we log an error here?
```suggestion
      } catch (err) {
        console.error("Failed to parse request body:")
        console.log(err)
        resolve(undefined)
      }
```

overlays/holo-nixpkgs/hpos-holochain-api/src/index.js
```diff
@@ -109,18 +119,16 @@ app.get('/dashboard', async (req, res) => {
 
     res.status(200).send(dashboard)
   } catch (e) {
-    return res.status(501).send(`hpos-holochain-api error: ${e}`)
+    return res.status(501).send(`hpos-holochain-api error: ${JSON.stringify(e)}`)
```
timotree3, line 122: Same thing with `util.inspect`

overlays/holo-nixpkgs/hpos-holochain-api/src/index.js
```diff
@@ -176,10 +184,10 @@ app.post('/install_hosted_happ', async (req, res) => {
       // Note: Do not need to install UI's for hosted happ
       return res.status(200).send(`Successfully installed happ_id: ${happId}`)
     } catch (e) {
-      return res.status(501).send(`hpos-holochain-api error: Failed to install hosted Happ with error - ${e}`)
+      return res.status(501).send(`hpos-holochain-api error: Failed to install hosted Happ with error - ${JSON.stringify(e)}`)
```
timotree3, line 187: `util.inspect(e)`

Connoropolous:
Thanks Alastair, I dont 100% follow you though. Can I request someone with the right system configuration for properly and fully running all the tests including these ones run them for this PR, so it can be approved with confidence?

timotree3 (approved):
Thanks so much! The fix + tests look great and I see that you cleaned up a bunch of other stuff too. I have a few small suggestions, but they're all not a big deal so this could just get merged as is.

Connoropolous:
@timotree3 i had a realization that you had actually added a commit that allows no data to be passed in validly to the hosted_happs endpoint. I think I undid that in this PR, thinking there was a mistake. A test case should he added that covers the no data passed scenario, and declared as 200 status, if thats indeed the intended behaviour

Connoropolous:
I see this is made irrelevant by the changes in #770 which touch this exact code and convert from body requests to search query arguments which is the better legit solution. Closing...

PR #768, opened 2021-03-30
hpos-configure-holochain bump

zo-el:


peeech (changes requested):
Unfortunately commit 33a0015 is breaking build. Multiple components of HPOS depend on version of rust nightly. So bumping nightly 1.5 years forward will break those components. 

BTW Hydra checks passed because commit 3bfd4df on base branch broke our CI. I fixed it in 2743a6c and merged base branch. You should now see errors on Hydra.

PR #770, opened 2021-03-31
B 03628 Dashboard Bugs

robbiecarlton:


Connoropolous (commented):
@robbiecarlton on line 125 of the server file there is a  JSON.parse without a try/catch block, which can crash the server given a non json string, I fixed it in another PR but closed it since you changed most of the same code in this PR. The JSON.parse thing is still worth fixing though. Take a look at #767 to see what I'm talking about and what I did.

Connoropolous:
Also as I was looking at it I couldn't figure out whether you actually needed hosted_happs to be able to take no interval... like Tim wrote here: https://github.com/Holo-Host/holo-nixpkgs/commit/b64502bde442c72211b46b532676fa75ac17de1d

robbiecarlton:
Hey @Connoropolous!

>on line 125 of the server file there is a JSON.parse without a try/catch block, which can crash the server given a non json string, 

Good catch, I'll update.

> Also as I was looking at it I couldn't figure out whether you actually needed hosted_happs to be able to take no interval... like Tim wrote here: [b64502b](https://github.com/Holo-Host/holo-nixpkgs/commit/b64502bde442c72211b46b532676fa75ac17de1d)

Hmm, that change by itself would still break because there's a call to service logger in `getPresentedHapps` that requires the interval

We could add a default but I'm going to leave it for now so clients are forced to be explicit.

overlays/holo-nixpkgs/hpos-holochain-api/src/index.js
```diff
@@ -121,9 +121,7 @@ app.get('/dashboard', async (req, res) => {
 
 app.post('/install_hosted_happ', async (req, res) => {
   // Loading body
-  const data = await new Promise(resolve => req.on('data', (body) => {
-    resolve(JSON.parse(body.toString()))
-  }))
+  const data = await parseBodyData(req)
```
Connoropolous, line 124: you could make a comment that parseBodyData is known to be safe, thus no try/catch here

robbiecarlton:
closing in favor of this:
https://github.com/Holo-Host/holo-nixpkgs/pull/774

PR #774, opened 2021-04-02
B-03628 dashboard bugs

robbiecarlton:


overlays/holo-nixpkgs/hpos-holochain-api/src/index.js
```diff
@@ -58,13 +59,14 @@ const getPresentedHapps = async usageTimeInterval => {
...
 
 app.get('/hosted_happs', async (req, res) => {
-  const usageTimeInterval = await Promise.race([
-    new Promise(resolve => req.on('data', (body) => {
-      resolve(JSON.parse(body.toString()))
-    })), 
-    new Promise(resolve => setTimeout(() => resolve(undefined), 100))
-  ])
-  if (usageTimeInterval !== undefined && !isusageTimeInterval(usageTimeInterval)) return res.status(501).send('error from /hosted_happs: param provided is not an object')
+  const usageTimeInterval = {
```
zo-el, line 62: Why are we switching from using the body to pass values?

zo-el (commented):
Just had a question about why we need to switch to using query?

Also if we have to
- we should also update the `/install-hosting-happ` endpoint
- we should also update `hpos-holochain-client` with a `dashboad` endpoint and if any changes that are needed because of the change in this PR

robbiecarlton:
> Just had a question about why we need to switch to using query?
> 
> Also if we have to
> 
> * we should also update the `/install-hosting-happ` endpoint
> * we should also update `hpos-holochain-client` with a `dashboad` endpoint and if any changes that are needed because of the change in this PR

@zo-el - I changed the two `get` end points to use query rather than body because you're not really supposed to use body with get requests. We don't need to change `install-hosting-happ` because it's a `post` endpoint.

Yes, I'll update `hpos-holochain-client`. I've barely used that tool, and had forgotten it exists.

timotree3:
`hpos-holochain-client` calls hosted_happs with no arguments, which is why I added the default value for the no body case. Could you update it to pass arguments? Maybe just some sensible default values since we are usually using it just to check that the right set of hosted_happs are present.

alastairong:
@robbiecarlton 

1. The sha256 for host-console-ui hasn't been updated. 
2. Please update the holo-nixpkgs integration tests accordingly as they currently just call /hosted-happs without any query params
3. Do we also need to update `configure-holochain`? I think it uses /hosted-happs endpoint as part of the install script? @zo-el @timotree3

alastairong (approved):
Reviewed and tested on my holoport. Works great

PR #777, opened 2021-04-08
Joining code 04 08- was 772, it will never die!!

zippy:
add in holo-hosting v0.1.0-alpha5 which includes ec 0.2.0-alpha3 which has the signing workaround


Br1ght0ne (approved):

peeech (approved):
This is massive PR that we should avoid in the future. Sub PRs have been reviewed and tests are passing, so I assume we are good to go.

PR #779, opened 2021-04-09
[DO NOTE MERGE] Proxy tx2 test

zippy:
bumps holochain for doing elemental-chat trycp testing with new proxy-server


alastairong:
@zippy Can we close?

PR #789, opened 2021-04-14
Nano Holochain testing

alastairong:


alastairong:
@peeech I'm going to keep this PR alive for now. When we get to a point when we're QAing hosted happs I will probably bring this back up to date so we can add Nanos to testing.

alastairong:
Closing as Holochain has caught up

PR #793, opened 2021-04-16
Reconfigure systemd services

peeech:
This PR is aiming to set correct order and properties of hosting-related systemd services on HPOS.

Required steps:
- [x] update systemd dependency tree
- [x] update services properties
- [x] ~~create `lair-shim` service?~~
- [x] analyze tests from perspective on new order

### DIFF
--- a/modules/services/configure-holochain.nix
+++ b/modules/services/configure-holochain.nix
@@ -26,8 +26,8 @@ in
 
   config = mkIf (cfg.enable) {
     systemd.services.configure-holochain = {
-      after = [ "network.target" "holochain.service" "hpos-holochain-api.service" ];
-      requisite = [ "holochain.service" "hpos-holochain-api.service" ];
+      after = [ "network.target" "hpos-holochain-api.service" ];
+      requisite = [ "hpos-holochain-api.service" ];
       wantedBy = [ "multi-user.target" ];
 
       environment.RUST_LOG = "configure_holochain=debug";
--- a/modules/services/configure-holochain.nix
+++ b/modules/services/configure-holochain.nix
@@ -44,7 +44,7 @@ in
         User = "configure-holochain";
         Group = "configure-holochain";
         ExecStart = "${cfg.package}/bin/configure-holochain ${cfg.working-directory}/config.yaml";
-        RemainAfterExit = true;
+        RemainAfterExit = true; # can be set to false because no other service depends on this one running
         StateDirectory = "configure-holochain";
         Type = "oneshot";
       };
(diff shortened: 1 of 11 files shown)


overlays/holo-nixpkgs/holochain/versions.nix
```diff
@@ -1,8 +1,8 @@
 {
   hpos = {
-    rev = "358018b1ae3c413bd2b58c0b12752c7b38779c09";
-    sha256 = "1af2vchilwdky6fjxvn93gr6ki793kwczry8nxm74qcikkx6rwb7";
-    cargoSha256 = "1nzr98dq8vg4av7p6ff19922k0mlpzcdzvrhwvdvbywawvawadx7";
+    rev = "12dd1d51dfae597b1f89ca1bbf57c475ff971427";
```
zo-el, line 3: What is this pointing to?

modules/services/configure-holochain.nix
```diff
@@ -26,8 +26,8 @@ in
 
   config = mkIf (cfg.enable) {
     systemd.services.configure-holochain = {
-      after = [ "network.target" "holochain.service" "hpos-holochain-api.service" ];
```
zo-el, line 29: What is going to trigger `configure-holochain` to start up after `holochain.service`?

overlays/holo-nixpkgs/holochain/versions.nix
peeech: To [this](https://github.com/holochain/holochain/pull/739) PR

modules/services/configure-holochain.nix
peeech: Ready state of `hpos-holochain-api.service` unit.

peeech:
Unfortunately `develop` of holochain is not compiling on HPOS because of an error in using defaut features. I need to fix it first :-(

modules/services/configure-holochain.nix
zo-el: Perfect

peeech:
> Unfortunately `develop` of holochain is not compiling on HPOS because of an error in using defaut features. I need to fix it first :-(

Fixed in https://github.com/holochain/holochain/pull/742

alastairong:
Looks great. I don't think I have ever seen a failure happen because Holochain service wasn't ready, but it's an important resiliency upgrade nonetheless.

peeech:
Merging this PR requires deciding which `holochain` revision are we gonna use

peeech:
@alastairong this one needs your approval as a code owner

alastairong:
I love that we have a second set of tests.

alastairong (approved):
Nice and tidy!

PR #794, opened 2021-04-20
Integrate host console stats

alastairong:


alastairong:
@robbiecarlton I've bumped to latest `develop` of both `holo-envoy` and `host-console-ui`

peeech (approved):

PR #795, opened 2021-04-21
Bump holochain to version including graceful genesis fail

peeech:


alastairong:
Closing to make something brand new off of `develop`

PR #796, opened 2021-04-23
Creates holochain-proxy service and profile

peeech:
Code below makes it possible to quickly and reliably create `holochain-proxy` instance by following instructions in RFCs repo under `devOps/holochain-proxy-setup`. Created instance can follow hydra channel of a choice therefore creating reliable and declaratively created environment.


peeech:
This PR is ready to merge into develop. Once merged all the Holoports that follow develop will communicate with proxy under IP address `45.55.107.33`, which is Floating IP. ATM this IP has a new server running `holochain-proxy.service` on NixOs machine. If things don't work as expected it is very easy to switch server behind `45.55.107.33` to old proxy (In [digitalocean's console](https://cloud.digitalocean.com/networking/floating_ips?i=d4f0d7).

profiles/logical/holo/holochain-proxy/default.nix
```diff
@@ -0,0 +1,21 @@
...
+  networking.hostName = "holochain-proxy";
+
+  services.holochain-proxy = {
+    enable = true;
+    cert-file = "proxy.cert";
+    working-directory = "/var/lib/holochain-proxy";
+    port = 5779;
+  };
+
+  services.openssh.enable = true;
```
alastairong, line 13: It doesn't look like you've opened port 80 for SSH?

profiles/logical/holo/holochain-proxy/default.nix
```diff
@@ -0,0 +1,21 @@
+{
+  imports = [ ../. ];
+
+  networking.hostName = "holochain-proxy";
+
+  services.holochain-proxy = {
+    enable = true;
+    cert-file = "proxy.cert";
```
alastairong, line 8: I assume you manually add the cert file?

modules/services/holochain-proxy.nix
```diff
@@ -0,0 +1,60 @@
...
+    cert-file = mkOption {
+      type = types.str;
+    };
+
+    working-directory = mkOption {
+      type = types.path;
+    };
+
+    package = mkOption {
+      default = pkgs.kitsune-p2p-proxy;
```
alastairong, line 26: Does this work without `with holochainAllBinariesWithDeps.hpos;` declared?

alastairong (approved):
@peeech I am pre-emptively reviewing this since my comments are just double-checking / for my own curiosity.

I assume the proxy is already deployed and has been confirmed as working?

profiles/logical/holo/holochain-proxy/default.nix
peeech: Setting `services.openssh.enable` to true opens port for ssh (22 BTW)

profiles/logical/holo/holochain-proxy/default.nix
peeech: Yep, details in [RFCs repo](https://github.com/Holo-Host/rfcs/blob/holochain-proxy-setup/devOps/holochain-proxy-setup.md#building-machine-from-scratch)

modules/services/holochain-proxy.nix
peeech: Yes, `kitsune-p2p-proxy` is defined [here](https://github.com/Holo-Host/holo-nixpkgs/blob/develop/overlays/holo-nixpkgs/default.nix#L159) as a part of `pkgs` (via overlays mechanism).

PR #798, opened 2021-04-24
update script for holochain

zo-el:


zippy (approved):
This looks great.  However, I didn't check to make sure this is idempotent.  I.e. if one of the git commands fails and we call `update-hc` again, will it still work?

Makefile
```diff
@@ -0,0 +1,53 @@
...
+# in the update-nix-by-failure will fail with exit code 102
+# and breaking it into 3 commands helps handle the error
+update-hc:
+	make HC_REV=$(HC_REV) update-hc-sha
+	make HC_REV=$(HC_REV) update-nix-by-failure
+	make HC_REV=$(HC_REV) update-hc-cargoSha
+	echo '⚙️  Creting new branch for release...'
+	git checkout -b update-hc-$(HC_REV)
+	git add overlays/holo-nixpkgs/holochain/
+	git commit -m overlays/holo-nixpkgs/holochain/version:hpos-rev:$(HC_REV)
```
zippy, line 19: It would be nice if the message only had the first 8 chars of the HC-rev

PR #801, opened 2021-04-26
auto-upgrade timer: startAt -> OnUnitActiveSec

### LINKED ISSUE #799: `holo-nixpkgs-auto-upgrade` servis creates Hydra overload
Each and every holoport in the world performs system upgrade check at the same moment. It creates request spikes on Hydra every 10 min. 

This issue can be easily avoided by switching from `OnCalendar` timer to `OnUnitActiveSec=10min` which will run every 10 min form system boot-up, that is creates random distribution of run times.

peeech:
fixes #799


alastairong (approved):
LGTM

PR #803, opened 2021-04-26
D-012420 fix regression and hpos-holochain-client bug

robbiecarlton:
This fixes a regression in `hpos-holochain-api` that was trying to read the body of `GET /hosted_happs` rather than getting the args from the query string.

It also fixes a bug in `hpos-holochain-client` where we were not properly passing the args in the query string.

### DIFF
--- a/overlays/holo-nixpkgs/hpos-holochain-client/hpos-holochain-client.py
+++ b/overlays/holo-nixpkgs/hpos-holochain-client/hpos-holochain-client.py
@@ -26,7 +26,7 @@ def hosted_happs(ctx, amount, duration_unit):
 @click.argument('duration_unit')
 @click.pass_context
 def dashboard(ctx):
-    print(request(ctx, 'GET', "/dashboard?duration_unit={duration_unit}&amount={amount}").text)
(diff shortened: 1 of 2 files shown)


overlays/holo-nixpkgs/hpos-holochain-client/hpos-holochain-client.py
```diff
@@ -19,12 +19,14 @@ def request(ctx, method, path, **kwargs):
 @click.argument('duration_unit')
 @click.pass_context
 def hosted_happs(ctx, amount, duration_unit):
-    print(request(ctx, 'GET', '/hosted_happs?duration_unit=${duration_unit}&amount=${amount}').text)
```
robbiecarlton, line 22: this is a JS string template, python was just passing the literal strings `'${duration_unit}'` and `'${amount}'` in the query

robbiecarlton:
> This looks good, but this changes hosted_happs such that now the usage time interval is required. Would it be better to make it optional like before?

We'd decided previously (the first time that code was merged) to make it required.

PR #808, opened 2021-04-29
[NO MERGE] Experiment/holochain darwin ldflags

steveeJ:


thedavidmeister (approved):

steveeJ:
@thedavidmeister I don't think you and I are in the position to approve this PR. The biggest counter indication to merging this is that it pulls in nixpkgs-unstable, which is clearly not desired for being the base for HPOS. Something that we can discuss is introducing a holochain overlay in this repository, which can have its own nixpkgs sources.

PR #811, opened 2021-04-29
D-012420 update sourcechain usage

robbiecarlton:
This updates servicelogger and hha to latest versions, fixes a small bug in `hpos-holochain-client` and includes holo-envoy from this branch:

https://github.com/Holo-Host/holo-envoy/pull/129

It shouldn't be merged until that PR is approved.


alastairong:
Rob posted that the source chain data is still wrong

alastairong (approved):
Tested as working and passing AC

PR #812, opened 2021-04-30
Use proxy-tx2 in develop

alastairong:
@zippy This uses tip of holochain develop. Please check that this is OK and includes the fix for the LMDB size.


alastairong:
@zippy 

I have pointed the new proxy-server to this channel and it is successfully running proxy-tx-2.
I have also pointed a holoport to it and used `proxy-tx2-cli` to confirm connectivity with the proxy server.

All seems to be in working order.

zippy (approved):

PR #814, opened 2021-05-03
README: complete binary cache public keys

steveeJ:


README.md
```diff
@@ -104,7 +105,8 @@ Otherwise, add `https://cache.holo.host/` to `substituters` and
 
 ```
 substituters = https://cache.holo.host/ https://cache.nixos.org/
-trusted-public-keys = cache.holo.host-1:lNXIXtJgS9Iuw4Cu6X0HINLu9sTfcjEntnrgwMQIMcE= cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY=
+trusted-public-keys = cache.holo.host-1:lNXIXtJgS9Iuw4Cu6X0HINLu9sTfcjEntnrgwMQIMcE= cache.holo.host-2:ZJCkX3AUYZ8soxTLfTb60g+F3MkWD7hkH9y8CgqwhDQ="
```
peeech, line 108: Given `https://cache.nixos.org/` is listed as one of the `substituters` we probably still need a `cache.nixos.org-1` trusted-public-key.

peeech: @steveeJ if this PR is intending to update README.md only I will be happy to finish it and merge.

steveeJ: > Given `https://cache.nixos.org/` is listed as one of the `substituters` we probably still need a `cache.nixos.org-1` trusted-public-key.

That's a good point, although I'm confused as to whether it's true. My system's I'm using don't seem to care whether or not I leave it on or not, as in they trust `cache.nixos.org` either way. This is probably because I'm using nix-daemon [which by default trusts cache.nixos.org](https://github.com/NixOS/nixpkgs/blob/1b688ca59ba78525308e963548eff6f3bebc221e/nixos/modules/services/misc/nix-daemon.nix#L333-L354). For system's that don't have a nix-daemon this might actually remove trust for it, although I'm not sure even after reading the Nix source code. I'll just add it back for now, but I want to note that there is risk of this key becoming deprecated eventually, which is why I was in favor of removing it.

I also changed the line above to use `extra-substituters` instead so we don't override the system's configuration which could divert from the default.

peeech (approved):
👍

PR #816, opened 2021-05-04
Resilient Hosted Happs

alastairong:
Includes
- Holochain tip of `sync-genesis` branch (`1050f6b`)
- Holo-Envoy tip of PR 124
- HHA v0.1.0-alpha10 (points to elemental chat v0.2.0-alpha7)

### DIFF
--- a/modules/services/holo-envoy.nix
+++ b/modules/services/holo-envoy.nix
@@ -29,7 +29,7 @@ in
       '';
 
       serviceConfig = {
-        Environment = "LOG_LEVEL=silly";
+        Environment = "LOG_LEVEL=debug";
         ExecStart = "${cfg.package}/bin/holo-envoy";
         TimeoutStartSec = 300;
         Restart = "always";
--- a/modules/services/holo-envoy.nix
+++ b/modules/services/holo-envoy.nix
@@ -29,7 +29,7 @@ in
       '';
 
       serviceConfig = {
-        Environment = "LOG_LEVEL=debug";
+        Environment = "LOG_LEVEL=silly";
         ExecStart = "${cfg.package}/bin/holo-envoy";
         TimeoutStartSec = 300;
         Restart = "always";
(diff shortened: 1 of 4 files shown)


profiles/logical/hpos/default.nix
```diff
@@ -232,7 +232,7 @@ in
       core_happs = [
        {
          app_id = "core-app";
-         bundle_url = "https://holo-host.github.io/holo-hosting-app-rsm/releases/downloads/v0.1.0-alpha8/core-app.0_1_0_alpha8.happ";
+         bundle_url = "https://holo-host.github.io/holo-hosting-app-rsm/releases/downloads/v0.1.0-alpha10/core-app.0_1_0_alpha10.happ";
```
zo-el, line 235: We should also update elemental-chat here for the self-hosted instance

modules/services/holo-envoy.nix
```diff
@@ -29,7 +29,7 @@ in
       '';
 
       serviceConfig = {
-        Environment = "LOG_LEVEL=debug";
+        Environment = "LOG_LEVEL=silly";
```
zo-el, line 32: This was intentionally set to debug because we do not need most of the silly logs and if you have found some of the silly logs useful we should update them to debug. 

There are a lot of silly logs that will just spam your logs

zo-el (changes requested):

profiles/logical/hpos/default.nix
alastairong: @zo-el self-hosted elemental chat was already bumped previously. This brings hosted elemental chat to the same version

modules/services/holo-envoy.nix
alastairong: @zo-el I'll change back to debug

alastairong:
```suggestion
        Environment = "LOG_LEVEL=debug";
```

zo-el (approved):
Looks Good

zippy (approved):

PR #822, opened 2021-05-09
Tweaks behavior of systemd services

peeech:
- [x] `hpos-holochian-api` does not depend on `holochain.service` (which caused errors during boot-up sequence)
 - [x] `hp-admin-crypto-server` always restarts
 - [x] `hpos-admin-api` always restarts


alastairong (approved):
This seems wise.

PR #831, opened 2021-05-12
HPOS resilient to holochain crashes and sign-up/sign-in issues

alastairong:
This updates `staging` with crashless holochain branch (holochain deactivates cells rather than panics)

Also, it merges in current `develop` which as:
- bump scripts
- holochain systemd service configured properly for database size
- systemd restart improvements for less flakiness
- Codeowner changes


zippy (approved):
Please note that this uses branch sets the membrane proof for the self-hosted instance to be the read-only membrane proof.  The ec DNA here 0.2.0-alpha11 has membrane proof checking disabled, so this will be ignored, but I'm noting this here to indicate that when we bump the DNA to one where it is no longer ignored, then self-hosted will become read-only.

PR #835, opened 2021-05-13
Update match engine 

alastairong:
Adds:
- auto-update
- Post HDK version of holochain
- Latest HHA and servicelogger

### DIFF
--- a/profiles/logical/holo/match-server/default.nix
+++ b/profiles/logical/holo/match-server/default.nix
@@ -10,6 +10,8 @@ let
   holochainWorkingDir = "/var/lib/holochain-rsm";
 
   configureHolochainWorkingDir = "/var/lib/configure-holochain";
+
+  kitsuneAddress = "kitsune-proxy://f3gH2VMkJ4qvZJOXx0ccL_Zo5n-s_CnBjSzAsEHHDCA/kitsune-quic/h/45.55.107.33/p/5788/--";
 in
 
 {
@@ -17,18 +19,18 @@ in
     ../.
   ];
 
-  environment.systemPackages = [ hc-state git ];
+  environment.systemPackages = [ hc-state git hpos-update-cli holochain hc kitsune-p2p-proxy ];
 
   networking.firewall.allowedTCPPorts = [ 80 443 ];
 
   services.lair-keystore.enable = true;
 
-  services.holochain = {
+  services.holochain = lib.mkDefault {
     enable = true;
     working-directory = holochainWorkingDir;
     config = {
-      environment_path = "${holochainWorkingDir}/databases";
-      keystore_path = "${holochainWorkingDir}/lair-keystore";
+      environment_path = "${holochainWorkingDir}/databases_lmdb4";
+      keystore_path = "${holochainWorkingDir}/lair-shim";
       use_dangerous_test_keystore = false;
       admin_interfaces = [
         {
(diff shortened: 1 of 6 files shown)


profiles/logical/holo/match-server/default.nix
```diff
@@ -10,24 +10,28 @@ let
...
 
-  services.holochain = {
+  services.holo-envoy.enable = lib.mkDefault true; # Enabling because Holochain systemd service depends on it. We're not using it or lair-shim
+
+  services.holochain = lib.mkDefault {
     enable = true;
     working-directory = holochainWorkingDir;
     config = {
-      environment_path = "${holochainWorkingDir}/databases";
+      environment_path = "${holochainWorkingDir}/databases_lmdb4";
```
peeech, line 34: How are we gonna keep those in sync across profiles?

profiles/logical/holo/match-server/default.nix
```diff
@@ -59,18 +76,24 @@ in
...
+       {
+         app_id = "core-app";
+         bundle_url = "https://holo-host.github.io/holo-hosting-app-rsm/releases/downloads/0_1_0_alpha17/core-app.0_1_0_alpha17.happ";
+       }
+       {
+         app_id = "servicelogger";
+         bundle_url = "https://holo-host.github.io/servicelogger-rsm/releases/downloads/v0.1.0-alpha7/servicelogger.0_1_0-alpha7.happ";
+       }
+      ];
+    };
```
peeech, line 89: How are we gonna keep those in sync across profiles?

profiles/logical/holo/match-server/default.nix
```diff
@@ -39,17 +43,30 @@ in
...
           proxy_config = {
             type = "remote_proxy_client";
-            proxy_url = "kitsune-proxy://nFCWLsuRC0X31UMv8cJxioL-lBRFQ74UQAsb8qL4XyM/kitsune-quic/h/proxy.holochain.org/p/5775/--";
+            proxy_url = kitsuneAddress;
           };
         }];
+        tuning_params = {
+          gossip_loop_iteration_delay_ms = 1000; # Default was 10
+          default_notify_remote_agent_count = 5;
+          default_notify_timeout_ms = 1000;
```
zippy, line 61: I'm worried about these `tuning_params`  We should remove all of the ones that don't have a `# Default was...` commint. We also need to delete those non default overrides in the base profile.

zippy (changes requested):

profiles/logical/holo/match-server/default.nix
alastairong: Does the database name need to be kept in sync across profiles?

profiles/logical/holo/match-server/default.nix
alastairong: Yea maybe we should extract it so we can unify across profiles.

profiles/logical/holo/match-server/default.nix
alastairong: @zippy See [22af9ae](https://github.com/Holo-Host/holo-nixpkgs/pull/835/commits/22af9aec84b07371e3a78a3b3dae16a7828caf7c)

Is it ok to have missing params? Are these overrides with defaults?

profiles/logical/holo/match-server/default.nix
```diff
@@ -10,24 +10,28 @@ let
   holochainWorkingDir = "/var/lib/holochain-rsm";
 
   configureHolochainWorkingDir = "/var/lib/configure-holochain";
+
+  kitsuneAddress = "kitsune-proxy://f3gH2VMkJ4qvZJOXx0ccL_Zo5n-s_CnBjSzAsEHHDCA/kitsune-quic/h/45.55.107.33/p/5788/--";
```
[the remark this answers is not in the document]
alastairong: We should extract it elsewhere. But that should be it's own PR I think

zippy (approved):

PR #837, opened 2021-05-15
B-04253: holo-auto-installer

zo-el:


modules/services/holo-auto-pilot.nix
```diff
@@ -0,0 +1,64 @@
...
+        StateDirectory = "holo-auto-pilot";
+        Type = "oneshot"; # This should be updated to run every 5 mins
+      };
+    };
+
+    users.users.holo-auto-pilot = {
+      isSystemUser = true;
+      home = "${cfg.working-directory}";
+      # ensures directory is owned by user
+      createHome = true;
```
peeech, line 59:
```suggestion
      createHome = false;
```
`working-directory` in question is already used by `configure-holochain` user

modules/services/holo-auto-pilot.nix
```diff
@@ -0,0 +1,64 @@
...
+
+    working-directory = mkOption {
+      default = "";
+    };
+  };
+
+  config = mkIf (cfg.enable) {
+    systemd.services.holo-auto-pilot = {
+      after = [ "network.target" "holochain.service" "hpos-holochain-api.service" "configure-holochain.service" ];
+      requisite = [ "holochain.service" "hpos-holochain-api.service" "configure-holochain.service" ];
```
peeech, line 34:
```suggestion
      after = [ "network.target" "configure-holochain.service" ];
      requisite = [ "configure-holochain.service" ];
```
`configure-holochain` already depends on `holochain.service` and `hpos-holochain-api.service`

modules/services/holo-auto-pilot.nix
```diff
@@ -0,0 +1,64 @@
...
+
+      serviceConfig = {
+        User = "holo-auto-pilot";
+        Group = "holo-auto-pilot";
+        ExecStart = "${cfg.package}/bin/holo-auto-pilot ${cfg.working-directory}/config.yaml ${cfg.working-directory}/membrane-proofs.yaml";
+        RemainAfterExit = true;
+        StateDirectory = "holo-auto-pilot";
+        Type = "oneshot"; # This should be updated to run every 5 mins
+      };
+    };
```
peeech, line 53:
```suggestion
        Type = "oneshot"; 
      };
      
      systemd.timers.holo-auto-pilot = {
        enable = true;
        wantedBy = [ "multi-user.target" ];

        timerConfig = {
          OnUnitActiveSec = "5min"; # run every 5 min
          OnBootSec = "2min"; # first run 2 min after boot
          Unit = "holo-auto-pilot.service";
        };
      };
    };
```

peeech: Timer will start `holo-auto-pilot.service` every 5 min. Please check the syntax before commiting, editing code in github UI seems very error-prone.

modules/services/holo-auto-pilot.nix
```diff
@@ -0,0 +1,64 @@
...
+      wantedBy = [ "multi-user.target" ];
+
+      environment.RUST_LOG = "holo_auto_pilot=debug";
+      environment.PUBKEY_PATH = "${cfg.working-directory}/agent_key.pub";
+      path = with pkgs; [ unzip ];
+
+      preStart = ''
+        sleep 2 # wait for configure-holochain to build config files
+      '';
+
```
peeech, line 44:
```suggestion
```
config files are built in `preStart` phase of `configure-holochain.service`. Because `holo-auto-pilot.service` depends on `configure-holochain.service` config files are in place by the time `holo-auto-pilot.service` runs, I think.

modules/services/holo-auto-pilot.nix
```diff
@@ -0,0 +1,64 @@
...
+  options.services.holo-auto-pilot = {
+    enable = mkEnableOption "holo-auto-pilot";
+
+    install-list = mkOption {
+      type = types.attrs;
+    };
+
+    membrane-proofs = mkOption {
+      type = types.attrs;
+    };
```
peeech, line 19:
```suggestion
```
Seems like you are only using `package`, `enable` and `working-directory`.

peeech (changes requested):

alastairong:
This works (tested across 2 holoports) and has the intended result of having a single happ_id.

I did notice that sometimes autopilot did not exit on completion (I called it in CLI because it wasn't on a timer). I had to ctrl+c to get back to prompt, but when I checked the logs it had finished.

modules/services/holo-auto-pilot.nix
zo-el: ok, tks :)

modules/services/holo-auto-pilot.nix
zo-el: cleaning up :)

alastairong (approved):

peeech (approved):
Approving, just one question. Why are we disabling here `systemd-resilience-test`?

PR #853, opened 2021-05-27
Auto pilot enable happ

alastairong:


overlays/holo-nixpkgs/hpos-holochain-api/tests/config.yaml
```diff
@@ -1,6 +1,10 @@
 core_happs:
   # HoloHosting App
+<<<<<<< HEAD
+  - bundle_url: https://holo-host.github.io/holo-hosting-app-rsm/releases/downloads/0_1_0_alpha17/core-app.0_1_0_alpha17.happ
+=======
   - bundle_url: https://holo-host.github.io/holo-hosting-app-rsm/releases/downloads/0_1_0_alpha16/core-app.0_1_0_alpha16.happ
+>>>>>>> auto-pilot-05-15
```
peeech, line 7: There's a merge conflict artifact here. Which one is correct?

profiles/logical/hpos/default.nix
```diff
@@ -43,7 +43,7 @@ in
   # REVIEW: `true` breaks gtk+ builds (cairo dependency)
   environment.noXlibs = false;
 
-  environment.systemPackages = with holochainAllBinariesWithDeps.hpos; [ git hc-state hpos-admin-client hpos-holochain-client hpos-reset hpos-update-cli holochain hc kitsune-p2p-proxy ];
+  environment.systemPackages = with holochainAllBinariesWithDeps.hpos; [ git hc-state hpos-admin-client hpos-holochain-client hpos-reset hpos-update-cli holochain hc kitsune-p2p-proxy hpos-config ];
```
peeech, line 46: This adds `hpos-config` to PATH in bash. Is that what we need?

peeech (changes requested):

overlays/holo-nixpkgs/hpos-holochain-api/src/index.js
```diff
@@ -194,6 +215,8 @@ app.post('/install_hosted_happ', async (req, res) => {
       } else {
         const serviceloggerPref = parsePreferences(preferences, happBundleDetails.provider_pubkey)
         console.log('Parsed Preferences: ', serviceloggerPref)
+
+        await enableHapp(happBundleDetails.happ_id)
```
zo-el, line 219: We should probably enableHapp only after the happ is properly installed(`installHostedHapp` is succesfull). Because currently enabling a happy would eventually update the KV store, but if the happ is not properly installed this can be an issue.

unless we also call deactivate happ on failure to install in the next lines, we should make the zome call to enable after we call `installHostedHapp`

zo-el (changes requested):

profiles/logical/hpos/default.nix
alastairong: Oh I just had it there to check the usage. I can remove it now

overlays/holo-nixpkgs/hpos-holochain-api/src/index.js
alastairong: Oh great point. Will update

overlays/holo-nixpkgs/hpos-holochain-api/tests/config.yaml
alastairong: alpha17

alastairong: Will update

overlays/holo-nixpkgs/hpos-holochain-api/src/index.js
alastairong: @zo-el Please take a look at my change. I want to keep everything self-contained to `hpos-holochain-api` so we don't have to revert changes in configure-holochain / auto-installer down the line.

profiles/logical/hpos/default.nix
alastairong: Addressed

overlays/holo-nixpkgs/hpos-holochain-api/tests/config.yaml
```diff
@@ -1,6 +1,10 @@
 core_happs:
   # HoloHosting App
+<<<<<<< HEAD
```
zo-el, line 3: @alastairong, I think you may have missed this

alastairong: Yes I think so too!

alastairong:
@zo-el Addressed issues

overlays/holo-nixpkgs/hpos-holochain-api/src/index.js
```diff
@@ -194,7 +215,9 @@ app.post('/install_hosted_happ', async (req, res) => {
       } else {
         const serviceloggerPref = parsePreferences(preferences, happBundleDetails.provider_pubkey)
         console.log('Parsed Preferences: ', serviceloggerPref)
-        await installHostedHapp(happBundleDetails.happ_id, happBundleDetails.happ_bundle.bundle_url, hostPubKey, serviceloggerPref, data.membrane_proofs)
+
+        let installed = await installHostedHapp(happBundleDetails.happ_id, happBundleDetails.happ_bundle.bundle_url, hostPubKey, serviceloggerPref, data.membrane_proofs)
+        if (installed) { await enableHapp(happBundleDetails.happ_id) }
```
zo-el, line 220: We don't really think this condition here,

zo-el (approved):
Just one clean up suggestion, else looks good

overlays/holo-nixpkgs/hpos-holochain-api/src/index.js
alastairong: What do you mean? I thought this is the logic to ensure we only call enableHapp if installHostedHapp succeeds

alastairong: @zo-el Or we could instead get the result of `listInstalledHapps` and call enable happ on each one? It's an idempotent zome call.

zo-el: I mean you do not need to check `if (installed)` because if you `await installHostedHapp` returns an error it will just fall into the try-catch. Else it will continue to enable the happ. 

So` installed` would never be false.

zo-el, line 220:
```suggestion
        await installHostedHapp(happBundleDetails.happ_id, happBundleDetails.happ_bundle.bundle_url, hostPubKey, serviceloggerPref, data.membrane_proofs)
        await enableHapp(happBundleDetails.happ_id)
```

peeech (approved):
Looks good

PR #856, opened 2021-06-01
Bump holochain that uses wasmer updates

zo-el:
Prepare holo-nixpkgs for the next round of testing.
- [x] Bumped holochain
- [x] Update rust 1.48 -> 1.52
- [x] Updated configure-holochain 
- [x] Update holo-auto-installer
- [x] Bump DNAs: EC, SL, HHA


alastairong (changes requested):
We should wait until we have a version of holochain that also has sqlite and bloom.

alastairong (approved):

PR #857, opened 2021-06-06
trycp with lair-shim

zo-el:


overlays/holo-nixpkgs/lair-shim-cli/default.nix
```diff
@@ -0,0 +1,22 @@
...
+
+    packageJSON = "${src}/package.json";
+    yarnLock = "${src}/yarn.lock";
+  
+    buildPhase = ''
+      ls 
+      cp -r ./deps/lair-shim/src $out
+    '';
+  };    
+}
```
timotree3, line 22: I like trailing newlines
```suggestion
}

```

modules/services/trycp-server.nix
```diff
@@ -24,7 +29,7 @@ in
 
       path = [ pkgs.holochain pkgs.lair-keystore ];
 
-      serviceConfig.ExecStart = "${cfg.package}/bin/trycp_server -r 9100-9400";
+      serviceConfig.ExecStart = "${cfg.package}/bin/trycp_server ${cfg.flags}";
```
timotree3, line 32: How would you pass a `--lair-shim` for testing? By editing this file?

timotree3 (approved):

modules/services/trycp-server.nix
alastairong: We generally enable lair-shim manually in configuration.nix, so I'd do it there so (excerpt from my holoport's file):

```
{ lib, ... }:

{
  imports = [
    <holo-nixpkgs/profiles/targets/holoport-plus>
    ./hardware-configuration.nix
  ];
  services.trycp-server.enable= true;
  services.trycp-server.flags = "--lair-shim 10000";
```

alastairong (approved):

PR #867, opened 2021-06-17
Auto pilot enable happ suggestion

zo-el:
…enabled


alastairong (approved):
I love this approach :)

PR #872, opened 2021-06-21
Prototype for holo host ui 

zo-el:


peeech (approved):
This works as expected

PR #874, opened 2021-06-21
Update branch protection settings

peeech:


.github/settings.yml
```diff
@@ -14,7 +14,7 @@ repository:
 collaborators: []
 
 teams:
-  - name: central
+  - name: holo-hosting
     permission: push
   - name: happy
```
zo-el, line 19: Just for clean up, should we remove this too?

zo-el (approved):
Just a note we have Paul d'Aoust and Stefan Junker in the holo-hosting team, and they get this permission too.

alastairong (approved):

.github/settings.yml
peeech: Good point. My reasoning is that happy still does exist on github, so if for any reason it is re-purposed they would definitely need to be able to push to this repo. I am  trying to reflect a current status on github. I am also gonna  send a note to TechOps that we are actively using github teams so that they should be careful with changes, because they are affecting our branch protection (therefore security model).

PR #878, opened 2021-06-22
[TEST] Hosted happs rc1

alastairong:


overlays/holo-nixpkgs/configure-holochain/default.nix
```diff
@@ -7,11 +7,11 @@ rustPlatform.buildRustPackage {
   src = fetchFromGitHub {
     owner = "Holo-Host";
     repo = "hpos-configure-holochain";
-    rev = "31a339e961bbe7b9177f3a721b5d29801c0d274b";
```
zippy, line 10: I't a little weird that this is pointing to just a plain branch with no PR.  Shouldn't we get this merged to develop of hpos-configure-holochain?

zippy (approved):
Looks ok, but I would prefer in the future that sha's of dependent repos  like holo-auto-installer, and hpos-configure-holochain weren't on dangling branches that haven't been through PRs and merge to develop for CI running purposes.

modules/services/holochain.nix
```diff
@@ -27,6 +27,10 @@ in
     working-directory = mkOption {
       type = types.path;
     };
+
+    restart-interval = mkOption {
+      type = types.str;
+    }
```
peeech, line 33:
```suggestion
    };
```
; missing

peeech (changes requested):

overlays/holo-nixpkgs/configure-holochain/default.nix
alastairong: `develop` of `holo-nixpkgs` and all related repos e.g. `hpos-configure-holochain` are currently on holochain version `24ceb63bdea374d1936b723e1966caf2e55ebfdc` which is LMDB + Bloom gossip, so it's technically ahead.

We could move our entire `develop` back but is it necessary?

PR #883, opened 2021-06-28
Matching service upgrades for ops console requirements

alastairong:
Adds:
- test-hp-manager to overlays: all source code was copied over from the repo because holo-nixpkgs doesn't want to build using fetchFromGithub
- ssh-pinger.service running every 2 hours at 30 past
- match-server ssh keys to holoCentral allowList
- mistake in matching-engine overlay that made a bin with the wrong script path
- an update to matching-engine overlay for a daily-uptime calculator service requested by ops console
- daily-uptime-calculator service that calculates a daily uptime summary in the format that ops console wants (runs daily)

Note: while we have a rebooter and channelSwitcher script added here, they aren't being used. Instead, we'll run them directly from our own computers as needed.


overlays/holo-nixpkgs/test-hp-manager/src/database.js
```diff
@@ -0,0 +1,39 @@
...
+  let data = fs.readFileSync(`${argv.configPath}`)
+  let credentials = await JSON.parse(data)
+  const username = credentials.MONGO_USERNAME
+  const password = credentials.MONGO_PASSWORD
+  const dbName = credentials.MONGO_DBNAME
+
+  // Connection URL
+  const url = `mongodb+srv://${username}:${password}@cluster0.hjwna.mongodb.net/${dbName}?retryWrites=true&w=majority`
+
+  // Open db connection and store it in global db (instanton pattern)
```
peeech, line 26:
```suggestion
  // Open db connection and store it in global db (singleton pattern)
```
Sorry for a mistake

overlays/holo-nixpkgs/test-hp-manager/src/data-handler.js
```diff
@@ -0,0 +1,68 @@
...
+//         $toDate: "$timestamp"
+//       }, 
+//       success: "$success" 
+//       }
+//     }
+//  }
+
+//  {
+//   timestamp: 1
+// }
```
peeech, line 68:
```suggestion
```

zippy (approved):
I don't see any blocking issues here, however note that I'm not deep into this code.

peeech (approved):

PR #894, opened 2021-07-08
Force enable SSH in master

alastairong:
Adding a mkForce in case users have conflicting ssh declarations in their local `configuration.nix`, which can happen due to customer support guidance

### DIFF
--- a/profiles/logical/hpos/default.nix
+++ b/profiles/logical/hpos/default.nix
@@ -284,7 +284,7 @@ in
 
   users.users.root.hashedPassword = "*";
 
-  profiles.development = {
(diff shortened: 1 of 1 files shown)


zippy (approved):

timotree3 (approved):
Not sure what mkForce does! but I heard you need another approval

alastairong:
@timotree3 It lets this declaration override any conflicting declarations elsewhere. Important in this case since users were guided during customer service to put SSH configurations in their local configuration.nix file.

Specifically, options have priority values of 1000 (lowest = highest priority).
mkOverride allows you to set a custom number.
mkForce sets the value at 50.

PR #895, opened 2021-07-08
Force enable ssh in develop

alastairong:
To mirror #894


peeech:
#896 takes care of it

PR #903, opened 2021-07-27
bump lair-keystore and rust

steveeJ:


zippy (approved):

alastairong:
@zippy @steveeJ Is this version of lair compatible with the current (quite old) version of Holochain in `develop`?

alastairong (changes requested):
This PR fails tests because this version of lair is not compatible with the version of holochain in `develop`

steveeJ:
> This PR fails tests because this version of lair is not compatible with the version of holochain in `develop`

it's more tricky than that, because of the different holochain versions a single lair-keystore version can't satisfy all of them. we can start a version map for lair-keystore in the same way we do it for holochain, or better yet combine the two.

PR #904, opened 2021-07-28
holo-envoy: Update to latest develop

JettTech:
### Updates:
 - bumps holo-envoy to `133b58d9892cbcea44d5321df8d16959a865ac96`


timotree3:
Merging this would have a big impact on the versions of elemental chat that holo-nixpkgs develop is compatible with, right? It would go to just the latest one that has the right cap tokens. That might be a blocker

PR #905, opened 2021-07-30
Infrastructure v4

zo-el:
# Changes:
- new Config V2
- Host holochain pub-key is the holoport ID

### TODO before merging:
 - gateway has to be manually deployed
 - registry will get auto deployed


peeech:
As mentioned in PR comment after merging this PR into `develop` we need to manually re-deploy devNet's `holo-router-gateway`. Router's functionality or API is not changing, so by upgrading devNet's `holo-router-gateway` we are not introducing braking changes into devNet.

alastairong (approved):

PR #913, opened 2021-08-17
proxy: revert port to int and convert elsewhere

alastairong:
Turns out it needed to be an int for some other option, so the easiest fix is to revert my previous change and convert it to string elsewhere.


JettTech (approved):
Looks good afaict from this PR.  I also agree that we should add integration tests to the CI to avoid issues like this in the future.

alastairong:
Tested this time :)

PR #914, opened 2021-08-17
Scaletest holofuel

timotree3:


profiles/logical/hpos/default.nix
```diff
@@ -318,4 +318,7 @@ in
   users.users.holo.isNormalUser = true;
 
   users.users.root.hashedPassword = "*";
+
+  services.trycp-server.enable = true;
+  services.trycp-server.flags = "-l 500";
```
alastairong, line 323: @timotree3 I think life will be easier if you put a `lib.mkForce`  in front of the value. Otherwise we'll have to manually disable / adjust holoport configurations when switching back and forth from this branch.

PR #923, opened 2021-09-06
Use fetchFromGitHub instead of local code

peeech:


alastairong (approved):
Looks good. Have you checked that it works? :)

peeech:
This PR introduces breaking changes to mongoDB structure. Should not be merged without the plan on how to handle those changes across all data users.

peeech:
> Looks good. Have you checked that it works? :)

Si. It works.

alastairong:
> This PR introduces breaking changes to mongoDB structure. Should not be merged without the plan on how to handle those changes across all data users.

@peeech I checked with Mary. We don't need a full migration of old data. It's sufficient to have it run for about 2-3 days to build up a history before switching over.

You need to coordinate with Matthew (Nichols) though to make sure that ops console updates their ETL processes accordingly.

PR #926, opened 2021-09-14
holo-registration-server: update jcf dna

zo-el:


overlays/holo-nixpkgs/holo-registration-service/src/index.js
```diff
@@ -36,31 +36,6 @@ app.get("/db-status", (req, res) => {
...
-    } catch (e) {
-      console.error("generate-code error: ", e);
-      return res.status(501).send(`hpos-holochain-api error: ${e}`)
-    }
-
-  } else {
-    return res.status(501).send(`hpos-holochain-api error: Failed to pass url in body`)
-  }
-})
-
```
alastairong, line 63: Why don't we need this endpoint anymore?

zo-el: I had it because I thought we could use it for tests, but there is a better way to create codes for tests. I have implemented it in HHA unit tests.

Also was a security hole.

PR #927, opened 2021-09-14
Infra V4: install core-app with right mem-proof

zo-el:


modules/services/configure-holochain.nix
```diff
@@ -11,17 +11,13 @@ in
     enable = mkEnableOption "configure-holochain";
 
     holo-network = mkOption {
```
peeech, line 13: Add comment describing different behavior of `configure-holochain` with `alphaNet` value.

profiles/logical/holo/holo-registration-server/default.nix
```diff
@@ -105,20 +110,13 @@ in
   services.configure-holochain = {
     enable = true;
     working-directory = configureHolochainWorkingDir;
+    holo-network = "registrationNet";
```
peeech, line 114:
```suggestion
    # on holo-regitration-server the value of holoNetwork should never be `alphaNet`
    holo-network = "dummyNet";
```

profiles/logical/holo/match-server/default.nix
```diff
@@ -74,28 +74,20 @@ in
   services.configure-holochain = {
     enable = true;
     working-directory = configureHolochainWorkingDir;
-    holo-network = settings.holoNetwork;
+    holo-network = "matchingNet";
```
peeech, line 77:
```suggestion
    # on holo-regitration-server the value of holoNetwork should never be `alphaNet`
    holo-network = "dummyNet";
```

profiles/logical/holo/holo-registration-server/default.nix
```diff
@@ -33,7 +33,12 @@ in
     email = "przemek.klimek@holo.host";
   };
 
-  services.lair-keystore.enable = true;
+  services.lair-keystore = {
+    enable = true;
+    preStart = "false";
```
peeech, line 38: Comment why would help here.

peeech (approved):
Just a request for commenting a code, otherwise great!!!

PR #935, opened 2021-09-20
RC1 Clean-up

alastairong:
Merge SSH pinger and match-service / match-engine fixes from release candidate


peeech:
This is fine. March server is following #923 which runs much newer version of a match server. But in fact API update is needed.

peeech (approved):
👍

PR #939, opened 2021-09-22
Add match-server SSH keys

alastairong:
Fixes alpha program issue with getting SSH uptime status


peeech (approved):
I'd prefer if we started issuing `ssh-ed25519` certs :-)

zo-el (approved):

PR #946, opened 2021-10-06
Test performance of EC on SSD-HP+

peeech:


peeech:
commit [30039c4](https://github.com/Holo-Host/holo-nixpkgs/pull/946/commits/30039c41ef439593e520a2cdf05d9aec7c7a5fda) is the only one relevant in this PR

PR #954, opened 2021-10-21
Bump holochain to v0.0.110

zo-el:


zo-el:
> Is it possible to bump this to 0.0.113 instead

Since we are making this a different ticket. Ill hold off

PR #976, opened 2021-11-16
[Infra 4]: Stabalize hpos-config file

zo-el:


profiles/logical/global-settings.nix
```diff
@@ -1,4 +1,5 @@
 {
   holoNetwork = "alphaNet";
   DEV_UID_OVERRIDE = "develop";
+  DEVICE_SEED_DEFAULT_PASSWORD = "pass";
```
peeech, line 4: If this pass is needed only by HPOS I would move it down to `profiles/logical/hpos/default.nix`

peeech (approved):
Great work, I'm impressed by the volume!

peeech:
I have tested entire registration flow on new infrastructure 4. It all works as expected, including failure cases and retries. Ready to merge.

PR #977, opened 2021-11-17
Hotfix: Deactivate servicelogger entirely

timotree3:


profiles/logical/hpos/default.nix
```diff
@@ -273,10 +273,6 @@ in
          app_id = "core-app";
          bundle_url = "https://holo-host.github.io/holo-hosting-app-rsm/releases/downloads/v0.1.1-alpha2/core-app-skip-proof.happ";
        }
-       {
-         app_id = "servicelogger";
-         bundle_url = "https://holo-host.github.io/servicelogger-rsm/releases/downloads/v0.1.0-alpha11/servicelogger.happ";
-       }
```
robbiecarlton, line 279: What's the value of removing this in addition to disabling above?

robbiecarlton (approved):

profiles/logical/hpos/default.nix
timotree3: The disable feature only disables the per hosted happ servicelogger, but does still passes through core-happs transparently

alastairong:
We merged #978 instead

alastairong:
@timotree3 Should we close this?

PR #981, opened 2021-11-18
Match server hotfix

alastairong:
- Bump max_hosts in KV store from 200 to 300
- Only keep last 14 days of uptime history
- Re-enable updating of DOMAIN2HAPP (disabled to allow for manually preventing users from accessing hApp)
- Fix bug where if 2+ happs were hosted, only hosts for 1 would be updated in HAPP2HOST


peeech (approved):
:+1:

PR #983, opened 2021-11-22
hydra: enable node exporter on both master and minion

cole-h:


profiles/logical/holo/hydra/default.nix
```diff
@@ -19,4 +19,47 @@
...
+          };
+        };
+      in
+      [
+        "--web.config=${config}"
+      ];
+
+    # See https://github.com/prometheus/node_exporter#collectors for a full list
+    # of supported collectors
+    enabledCollectors = [
```
cole-h, line 42: This list is mostly arbitrary -- happy to add or remove exporters as desired.

alastairong:
I think we're going to need a little context on this PR?

cole-h:
Sure! Please see https://github.com/Holo-Host/bugs-n-issues/issues/5 and the tickets that reference it.

profiles/logical/holo/hydra/default.nix
peeech: Yes, I'd love to discuss the list with you in details.

peeech: I am curious of your opinion @cole-h how useful are those collectors in our case.

Let's assume that our main goal is to optimize machine for building jobs in our specific context. I can see that some of collectors, like `vmstat`, `sytemd`, `diskstats` will give us meaningful insight into hardware resources utilization on hydra, others like `arp`, `entropy` will give us information that is much less of a value in our case.

The question here is if there's any negative consequence of collecting too much data? Apart from the size of the log file size I cannot really see any.

profiles/logical/holo/hydra/default.nix
```diff
@@ -19,4 +19,47 @@
...
+      "mdadm"
+      "meminfo"
+      "netdev"
+      "netstat"
+      "sockstat"
+      "systemd"
+      "textfile"
+      "time"
+      "vmstat"
+      "wifi"
```
peeech, line 62:
```suggestion
```
No wifi networking interface avail on our machines

profiles/logical/holo/hydra/default.nix
grahamc: In general there is not much consequence of collecting too much.

The machines running the node exporter don't store any of the data, so there is no local data use concern. The data we collect and store gets pruned after a while, and while we'll collect samples 4 times a minute, we can roll it up in to 1 sample per minute, or 1 sample per 5 minutes, etc. after the fact.

Ideally we'll be able to get this rolled out so we can start monitoring the systems. This list is not totally relevant everywhere we use it, but we haven't found a compelling reason to reduce it -- especially when sometimes a previously irrelevant-seeming collector suddenly proves useful while debugging a production issue.

profiles/logical/holo/hydra/default.nix
grahamc: Thanks, removed!

profiles/logical/holo/hydra/default.nix
peeech: perfect

PR #998, opened 2021-12-06
Bump to 0.0.118+ for gossip fixes

alastairong:


alastairong:
Closing this since it looks like we'll need more fixes for scalability anyway

PR #1004, opened 2021-12-07
overlays/mk-jobsets: prefix PR jobs with "pr-"

grahamc:
Right now PRs in Hydra are named based on their PR number, ie this PR gets number "1004". However, Hydra generally does not allow names to contain only numbers, so e.g.
attempting to manually evaluate one of these jobsets will make Hydra
error:

    09:54:04 hydra-server.1       | [error] Caught exception in Hydra::Controller::API->push "invalid jobset specifier ‘test2:2’"
    09:54:04 hydra-server.1       | [error] Caught exception in Hydra::Controller::Root->end "Wide character at /home/vin/workspace/vcs/hydra/src/script/../lib/Hydra/Controller/Root.pm line 283."

This PR prefixes each jobset with `pr-`, so this PR will be named `pr-1004` instead of `1004`.


peeech:
closes https://github.com/Holo-Host/bugs-n-issues/issues/13

PR #1010, opened 2021-12-12
Staging branch for updated Hydra

peeech:
In this PR a number of updates is included:
- [x] prefix PR jobs with "pr-"
- [x] enable node exporter on both master and minion
- [x] update to nixos-21.11

Because the context here is specifically for Hydra we need to stage-deploy this code to `hydra-backup` instance first and evaluate performance before merging to `hydra` branch.


profiles/logical/holo/hydra/master/default.nix
```diff
@@ -15,6 +15,7 @@ in
   nix.buildMachines = [
     {
       hostName = "localhost";
+      maxJobs = 4;
```
peeech, line 18: Interestingly enough I had a chance to compare hydra-backup that follows #1010 (`maxJobs = 4`) and hydra that follows `hydra-19.09` branch (`maxJobs = 2`). hydra-backup finished builds of #1010 twice as fast as hydra, which seems to be pointing to an issue of competition for resources at values of maxJobs higher than 2.
Also "hydra-minion-1.holo.host" with `maxJobs = 96` does seem to evaluate much slower than `localhost`, even though it is a much more powerful machine.

alastairong (dismissed):

alastairong (approved):

timotree3 (approved):

PR #1011, opened 2021-12-13
clean up: holo-registration-service -> membrane-proof-server

zo-el:


peeech (approved):
Looks good, I have not run this profile on my machine though.

PR #1015, opened 2021-12-14
Infra v4: trigger holo-auth onMount

zo-el:


zo-el:
I don't think we need this feature based on the recent hardening of the registration process.

PR #1017, opened 2021-12-15
Hosted happ cloning

robbiecarlton:


overlays/holo-nixpkgs/hpos-holochain-api/src/api.js
```diff
@@ -37,7 +38,7 @@ const installHostedHapp = async (
       agent_key: agentPubKey,
       installed_app_id: happId,
       membrane_proofs: membraneProofs || {},
-      uid: DEV_UID_OVERRIDE
+      uid: isClone ? `${DEV_UID_OVERRIDE}:${happId}` : DEV_UID_OVERRIDE
```
zo-el, line 41: I think this should be `${DEV_UID_OVERRIDE}${happId}` without the `:`, because if there is not a `DEV_UID_OVERRIDE` the uid would be `:happid` and if people want to join the same hash space it will be something that they would have to know.

zo-el (changes requested):
one small change request

zo-el (approved):
Looks good. The HHA may need to be updated with happy to include mem-proofs, But right now we are not worried about it. So it is good.

peeech:
@zo-el @robbiecarlton what's up with this PR? Are we still trying to merge it?

robbiecarlton:
> @zo-el @robbiecarlton what's up with this PR? Are we still trying to merge it?

@peeech - Yeah. Tomorrow I'm going to be cleaning up and finishing this ticket and will update and try and get this merged as part of that

zo-el:
Just merged this work into a working branch: #1245

PR #1022, opened 2021-12-17
Infra v4: direct registration email without challenge

zo-el:
> no challenge email required
## Work done:
- holo-auth here: https://github.com/Holo-Host/holo-auth/pull/46
- hpos-led-manager checks for zt-notification files that indicates that zt registration is complete


peeech (approved):
Code looks good. Because there is a substantial change in a registration flow introduced in this PR my request is to check this flow on holoports against a "Holo Registration and Setup" Miro board.
Also I would like to test if the following corner cases give expected outcome:
- user deletes `/var/lib/configure-holochain` folder only (without resetting entire holoport)
- user deletes `/var/lib`
- user plugs in USB with different credentials than ones used for obtaining mem-proof

PR #1025, opened 2021-12-20
Infra v4: mem-proof-server: return proof when exists

zo-el:
## Work done:

- mem-proof-server returns existing mem-proof for agents that are already registered on the ops-console-db
- configure-holochain starts after holo-auth


peeech (approved):
Code looks good, I am not sure if this code executes business logic as expected, because I have not read any.

PR #1026, opened 2021-12-21
Infra v4: LED turns RED when agent keys do not match

zo-el:
## Work done:

- led-manager checks if config agents and stored agents are the same. if it is not then the LED turns RED that notifies the host should be wiping their holoport before using the new config
- led-manager switched to using stable rust
- Make global vars that are needed for led-manager and holo-auth


profiles/logical/hpos/default.nix
```diff
@@ -113,7 +113,7 @@ in
   services.mingetty.autologinUser = "root";
 
   services.hpos-led-manager.kitsuneAddress = holoNetwork.proxy.kitsuneAddress;
-
+
```
peeech, line 116:
```suggestion

```

peeech: Dangling white space will trigger nix linter on hydra

peeech (approved):

PR #1027, opened 2021-12-22
Infra v4: admin-portal v0.1.5

zo-el:
## Work done:

- update admin portal to use the new version of key-manager
- Fix bugs in admin-portals signatures with payload


overlays/holo-nixpkgs/host-console-ui/default.nix
```diff
@@ -6,8 +6,8 @@
     src = fetchFromGitHub {
       owner = "holo-host";
       repo = "host-console-ui";
-      rev = "da009ce9e486ffa9e9312377064da886ac6d1a68";
-      sha256 = "12fvyqmpqi4xmq2x5ydb92bdnilwnvb5d5iykzzl0g8ak2w2arjp";
+      rev = "1905a33397ba0da85c24d0c3dbf92a1340df683b";
```
peeech, line 9: Hmm, you are pointng here to [revision](https://github.com/Holo-Host/host-console-ui/pull/35/commits/1905a33397ba0da85c24d0c3dbf92a1340df683b) that is not merged into master, nor it is passing tests. I think that both of those issues should be fixed before this PR is mergable

peeech (changes requested):
See comments

overlays/holo-nixpkgs/host-console-ui/default.nix
alastairong: @Holo-Host/holo-hosting has agreed to waive working agreement for this.

alastairong (approved):

PR #1031, opened 2022-01-06
Pass AUTH_SERVER_URL to holo-auth-client

peeech:
Auth server url is now defined in `holo-networks.nix` and passed to `holo-auth-client` service. As a consequence holoports connected to `devNet` will use `https://devnet-auth-server.holo.host` for registration process, while holoports connected to `alphaNet` will use `https://auth-server.holo.host`.

Todo:
- [x] add authServerUrl to holoNetwork option and make use in services
- [x] bump `holo-auth-client` to version that reads AUTH_SERVER_URL env var


peeech:
This is stale PR ATM

PR #1012, opened 2021-12-13
Auto update nixpkgs

peeech:
This PR, once merged into develop, brings in an auto-triggered script for bumping NixOs on holo-nixpkgs. Quick facts about script:
- it auto-generates PR against develop that bumps version of NixOs in `holo-nixpkgs` to latest stable
- it is triggered by cron on the first day of each month

Once PR is created we need to manually trigger nightly-tests with this PR as a parameter for holoports
Once PR is merged we need to make sure all the DevNet has updated to latest develop and run nightly-test again

PR #1032, opened 2022-01-06
Cleanups / fixes for offline installer

cole-h:
These commits are a part of the offline installer work (https://github.com/Holo-Host/bugs-n-issues/issues/15) and are necessary for the image to successfully build.

Note that this is based against the `auto-update-nixpkgs` branch (PR https://github.com/Holo-Host/holo-nixpkgs/pull/1012).


modules/services/configure-holochain.nix
```diff
@@ -55,8 +55,10 @@ in
       };
     };
 
+    users.groups.configure-holochain = { };
     users.users.configure-holochain = {
       isSystemUser = true;
+      group = "configure-holochain";
```
cole-h, line 61: Note that this is *different* from what was originally happening -- according to the docs, NixOS would default "no group" to the group `nogroup` (even if there was a group by the same name). Happy to change this back to what it was, functionally (`group = "nogroup";`) if this has potential to cause issues.

modules/services/hpos-led-manager.nix
```diff
@@ -23,6 +23,7 @@ in
 
     kitsuneAddress = mkOption {
       type = types.str;
+      default = "";
```
cole-h, line 26: Is there a better default that could be used here? If some outside consumer of this module forgets to set the `kitsuneAddress`, I'd imagine whatever uses it will cease to work due to it being an empty string. This is just the minimal viable change to get things building again.

alastairong: This is fine. It's actually not currently in use

PR #1033, opened 2022-01-06
Offline installation / recovery image

cole-h:
Closes https://github.com/Holo-Host/bugs-n-issues/issues/15.

Depends on https://github.com/Holo-Host/holo-nixpkgs/pull/1012 and https://github.com/Holo-Host/holo-nixpkgs/pull/1032 (and is based on them).

Marked as a draft because https://github.com/Holo-Host/holo-nixpkgs/pull/1032 should be merged first.

The most relevant commits are 510b74af97503952907ce678ab00e5f3f88021f0 and 8e06acc3a31bcaf9073ef9c82b543b64cd1e8e07 (although the rest are required to build successfully).

Also note that this PR is based against `auto-update-nixpkgs` (PR #1012) to incorporate its fixes.

---

I built the image with `nix-build nixpkgs/nixos -A config.system.build.isoImage -I nixos-config=./profiles/installers/holoport/default.nix` and booted the installer inside a qemu VM with `qemu-img create mydisk.img 50G && qemu-system-x86_64 -cpu host -enable-kvm -boot d -cdrom ./result/iso/hpos-for-holoport.iso -m 8192 -hda mydisk.img -nic none`.

If you're using a new enough version of Nix, it's possible to run into issues such as `fish: Job 1, 'nix-build nixpkgs/nixos -A conf…' terminated by signal SIGABRT (Abort)` -- if you do, prepend the `nix-build` with `GC_DONT_GC=1`. (FWIW, I didn't see this with `nix (Nix) 2.5.1`, but I did see it with `nix (Nix) 2.6.0pre20220105_6d78446`.)


overlays/holo-nixpkgs/hpos-install/default.nix
```diff
@@ -39,8 +40,12 @@ let
           mkpart primary 0% 100% \
           set 1 boot on
 
-        mkfs.ext4 -F /dev/sda1
+        mkfs.ext4 -F /dev/sda1 -L HOLOINSTALL
```
cole-h, line 43: These labels are arbitrary, but are necessary so that it is possible to identify the disk (as the `uuid` is different for every formatted disk).

profiles/installers/default.nix
```diff
@@ -29,6 +32,8 @@ in
 
   boot.postBootCommands = ''
     mkdir -p /mnt
+
+    ln -sf ${closure.config.system.build.toplevel} /built-closure
```
cole-h, line 36: `/built-closure` is an arbitrary location; it can be anywhere you like, but should be static.

One alternative approach that is possible would be to pass this path in as an argument of the auto-installer unit, which could be configured with something like `services.hpos-install.closure = closure.config.system.build.toplevel;` and passed to the `hpos-install` script.

overlays/holo-nixpkgs/hpos-install/hpos-install.sh
```diff
@@ -16,13 +16,14 @@ trap on_exit EXIT
...
 cat @configuration@ > /mnt/etc/nixos/configuration.nix
 
 nixos-install --channel @channel@ --no-root-passwd \
-  -I holo-nixpkgs=@channel@/holo-nixpkgs \
-  -I nixpkgs=@channel@/holo-nixpkgs/nixpkgs \
-  -I nixpkgs-overlays=@channel@/holo-nixpkgs/overlays
+  --system /built-closure
+
 echo '@channelUrl@ holo-nixpkgs' > /mnt/run/.nix-channels
+ln -sf /run/.nix-channels /mnt/root/.nix-channels
```
cole-h, line 27: This is necessary because of c84c2d01097806b10641f25d5154991f5b81288d -- `nix-channel` looks for the channels list in `$HOME/.nix-channels`, but that commit moved said list out of its expected location. Without this, the `holo-nixpkgs-auto-updater` unit would fail because `nix-channel --update` would fail on an empty file.

overlays/holo-nixpkgs/hpos-install/default.nix
```diff
@@ -39,8 +40,12 @@ let
...
         mount /dev/sda1 /mnt
+
+        ln -fs \
+          $(${util-linux}/bin/lsblk -no pkname /dev/sda1) \
+          /dev/inferred-grub
```
cole-h, line 48: This is necessary because this `extraPrepareConfig` stuff from https://github.com/Holo-Host/holo-nixpkgs/blob/c5cf777ba3225a55d6fff0c7b8c9daaef5dc23ae/profiles/physical/hpos/inferred-grub.nix#L6 doesn't get run until it's time to install grub, which was too late in my testing.

PR #1034, opened 2022-01-07
Infra v4: holo-auth-notifier

zo-el:


alastairong (approved):
Approving but I think we should document somewhere that we expect that if holoNetwork =/= devNet, we expect production holo-router (i.e. holohost.net).

This is just an edge case if a user tries to register a holoport on a non-standard network though.

PR #1039, opened 2022-01-12
profiles/logical/global-settings: switch to alphanet

zo-el:


alastairong:
Closing in favour of #1040

PR #1043, opened 2022-01-13
[B-04778] verify-registration-code endpoint

timotree3:
Changes
- Adds the endpoint
- Moves most of `index.js` into `app.js` because I was trying to make sure the server spawned in `index.js` wasn't preventing the tests from exiting. I ended up putting `--forceExit` so this change wasn't necessary, but I think it's fine to keep.
- Improve `setup-holochain.js` so it will be easier for future people to debug issues with holochain version.


overlays/holo-nixpkgs/membrane-proof-service/README.md
```diff
@@ -43,9 +43,39 @@ Node API service that verifies if a holo user is registered by Holo.ltd.
...
 ```
   {
-    error: String,  // "<short discription of Error>",
+    error: String,  // "<short description of Error>",
     info: String, // "<user understandable error>",
-    isDisplayedToUser: bool, //<"should be displayed in chaperone or not">
+    isDisplayedToUser: bool, // should be displayed in chaperone or not
+  }
+```
+- `/verify-registration-code/` (PUT): Ensures that the provided registration code is associated with the provided email. (Used in quickstart)
```
zo-el, line 51: Why is this a PUT? It should be a GET call

overlays/holo-nixpkgs/membrane-proof-service/src/helper/verify-code.js
```diff
@@ -0,0 +1,40 @@
...
+
+        let codeList = record.registrationCode.find(a => a.code == registration_code)
+        if (codeList == undefined) {
+            codeMissing = true
+            continue;
+        }
+        // TODO: Should we ensure `codeList.status === "sent"`?
+        console.log("Verification successful")
+        return res.status(200).send({});
+    }
```
zo-el, line 23: Let sync this code up with the `/register-user` endpoint. Let us make one function that both the endpoints can use to get a verified registration code, so that they stay in sync.
The function will:
- get record based on email
- validate it
- if valid return the record
- if invalid returns an error.

And then the endpoints can do what they need to with the result

zo-el (changes requested):

overlays/holo-nixpkgs/membrane-proof-service/README.md
timotree3: GET calls can't have bodies

overlays/holo-nixpkgs/membrane-proof-service/src/helper/verify-code.js
timotree3: Okay, I'll do that tomorrow. One issue I foresee is that in `verifyCode`, I am distinguishing between (email found but membrane proof not found) and (email not found), but that distinction is absent in `registerUser`. I think this distinction is meaningful for the quickstart UX, so I don't want to get rid of it. Should I just add the same distinction to `registerUser`?

overlays/holo-nixpkgs/membrane-proof-service/README.md
zo-el: Yes, you would have to pass `params` instead of a body

overlays/holo-nixpkgs/membrane-proof-service/src/helper/verify-code.js
zo-el: It is useful for quickstart, so let's keep it for the `verifiyCode`, but we already have set errors types that need to be sent by the `registerUser` endpoint, So we d oo not want to add that case in `registerUser`

zo-el (approved):
Looks good

PR #1047, opened 2022-01-14
Infra v4:  holo_agent_override based on holoNetwork

zo-el:
PR 1: ability to set .dna files instead of .happ and add properties: 
- https://github.com/Holo-Host/hpos-configure-holochain/pull/124
---
PR 2: 
- dependency:
  - https://github.com/Holo-Host/holo-auth/pull/53
- Led manage changes:
  - Check if the device_bundle is the same as the cached pub_key before it checks for zerotier check and tls check.
---
PR 3: 
- configure-holochain: updated to create random agent only in devNet
  - https://github.com/Holo-Host/hpos-configure-holochain/pull/125
- led-manager: compare device_bundle and the cached_pub_key  only in devNet


profiles/logical/holo/match-server/default.nix
```diff
@@ -40,6 +40,16 @@ in
        {
          app_id = "core-app";
          bundle_url = "https://holo-host.github.io/holo-hosting-app-rsm/releases/downloads/0_1_1_alpha7/core-app.0_1_1_alpha7.happ";
+         dnas = [
+           {
+             id = "core-app";
+             url = "https://holo-host.github.io/holo-hosting-app-rsm/releases/downloads/0_1_1_alpha7/holo-hosting-app.0_1_1_alpha7.dna";
+           }
+         ];
+         properties =''
```
zo-el, line 49: Comment about this needing to be a valid YAML file

profiles/logical/holo/match-server/default.nix
```diff
@@ -40,6 +40,16 @@ in
        {
          app_id = "core-app";
          bundle_url = "https://holo-host.github.io/holo-hosting-app-rsm/releases/downloads/0_1_1_alpha7/core-app.0_1_1_alpha7.happ";
```
zo-el, line 42: Comment:  that the bundle url is still required to get app-id

alastairong (approved):
Approved in live review with Joel. Requested comments about yaml formatting and low urgency code cleanup issue for future

peeech (approved):
I think that `hpos-led-manager` should be in a separate repo. Next step would be moving it out of `holo-nixpkgs`

Approved with comment above

PR #1053, opened 2022-01-17
Infra v4: holo-init triggeres reset

zo-el:


overlays/holo-nixpkgs/hpos-init/hpos_init/main.py
```diff
@@ -70,9 +70,21 @@ def on_wormhole_code(wormhole_code):
...
+    return glob('/media/*/hpos-reset')
+
+def check_reset():
+    reset_file = scan_reset_file()
+    if reset_file:
+        try:
+            os.system("rm -f /media/*/hpos-reset")
+            os.system("hpos-reset")
+        except:
+            print("unable to reset")
```
alastairong, line 83: @zo-el If there is an exception and it fails to reset, will it continue to load the config?

I think it should so that we still land on the red LED.

alastairong (approved):

PR #1063, opened 2022-01-18
overlays/holochain-nixpkgs: init

cole-h:
Addresses https://github.com/Holo-Host/bugs-n-issues/issues/21.

Based against https://github.com/Holo-Host/holo-nixpkgs/pull/1012 to pick up the commits that fix compatibility with newer Nixpkgs. Also contains a couple of commits from https://github.com/Holo-Host/holo-nixpkgs/pull/1032 that fix some other issues with newer Nixpkgs. This also allowed me to run `nix-build tests -A hpos-api-tests`, which succeeded.


overlays/holochain-nixpkgs/holochain/versions/hpos.nix
```diff
@@ -1,10 +1,10 @@
 # This file was generated with the following command:
-# update-holochain-versions --git-src=revision:b194f98a007a1dc9e34f7997cd6e6a33f65807f8 --lair-version-req=~0.0 --output-file=holochain_version.nix
+# update-holochain-versions --git-src=revision:holochain-0.0.117 --output-file=holochain_version.nix
 # For usage instructions please visit https://github.com/holochain/holochain-nixpkgs/#readme
 
 {
     url = "https://github.com/holochain/holochain";
-    rev = "b194f98a007a1dc9e34f7997cd6e6a33f65807f8";
+    rev = "holochain-0.0.117";
     sha256 = "sha256-O9VHSbnCFoC5FcL9uHWAUUgr23AEbNnerm1N/9vBuPc=";
```
cole-h, line 8: This is the same hash: https://github.com/Holo-Host/holo-nixpkgs/blob/ff0ddee477161769adab5ab551e249c78a6d35d7/overlays/holo-nixpkgs/holochain/versions.nix#L4

```
$ rnix-hashes '1xxqq7dzykbdmvgdjv04f3djnj2ih1svizf22nwq05n2p54lgm9v'
SRI     sha256-O9VHSbnCFoC5FcL9uHWAUUgr23AEbNnerm1N/9vBuPc=
base16  3bd54749b9c21680b915c2fdb8758051482bdb70046cd9deae6d4dffdbc1b8f7
base32  1xxqq7dzykbdmvgdjv04f3djnj2ih1svizf22nwq05n2p54lgm9v
base64  O9VHSbnCFoC5FcL9uHWAUUgr23AEbNnerm1N/9vBuPc=
```

overlays/holochain-nixpkgs/default.nix
```diff
@@ -0,0 +1,47 @@
...
+  };
+
+  holochainMeta = final.callPackage "${holochainNixpkgs}/packages/holochain" { };
+in
+{
+  inherit (holochainMeta)
+    # utility functions
+    mkHolochainBinary
+    mkHolochainAllBinaries
+    mkHolochainAllBinariesWithDeps
```
steveeJ, line 17: are tehse still in use anywhere in this repo? if not i'd prefer to not expose them through this repo

overlays/holochain-nixpkgs/default.nix
```diff
@@ -0,0 +1,47 @@
...
+    ;
+
+  holochain_main = final.holochainAllBinariesWithDeps.main.holochain;
+  hc_main = final.holochainAllBinariesWithDeps.main.hc;
+  lair-keystore_main = final.holochainAllBinariesWithDeps.main.lair-keystore;
+  kitsune-p2p-proxy_main = final.holochainAllBinariesWithDeps.main.kitsune-p2p-proxy;
+  holochain_develop = final.holochainAllBinariesWithDeps.develop.holochain;
+  hc_develop = final.holochainAllBinariesWithDeps.develop.hc;
+  lair-keystore_develop = final.holochainAllBinariesWithDeps.develop.lair-keystore;
+  kitsune-p2p-proxy_develop = final.holochainAllBinariesWithDeps.develop.kitsune-p2p-proxy;
```
steveeJ, line 46: if this repo doesn't use these we can remove them from here as holochain-nixpkgs already tracks these

overlays/holochain-nixpkgs/holochain/versions/hpos.nix
steveeJ: do you intend to keep the versions.nix around? i can't see it being removed in this PR

overlays/holochain-nixpkgs/default.nix
```diff
@@ -0,0 +1,47 @@
+final: prev:
+let
+  holochainNixpkgs = final.fetchFromGitHub {
```
steveeJ, line 3: in some other repos i've started to use `niv` for sources management. could be a consideration for this one as well.

steveeJ (commented):
this is a great start, thank you! i think there's still code that can be removed, which is my main motivation for this first iteration :-)

overlays/holochain-nixpkgs/default.nix
cole-h: I agree that `niv` would be much better, but I think that should be a follow-up PR. That way we could add `nixpkgs` and everything else (e.g. `mozilla-nixpkgs`) from the `holo-nixpkgs` overlay.

overlays/holochain-nixpkgs/default.nix
cole-h: I believe this is still necessary so that they are built and tested by Hydra (I think that's what https://github.com/Holo-Host/holo-nixpkgs/blob/2d904205aac0c25c0b9e3721aa43e0aa055ecffb/release.nix#L12 is doing).

steveeJ: the reason why i originally put these here is because users of [holonix](https://github.com/holochain/holonix) were instructed to use holo's hydra instance as a binary cache. this is no longer the case and so from my perspective there is no need to keep these around here nor building them on holo's hydra.

overlays/holo-nixpkgs/default.nix
```diff
@@ -305,4 +305,6 @@ rec {
       };
     }
   );
+
+  pkgconfig = pkg-config;
```
cole-h, line 309: I've fixed this locally, but this also requires a change to `holochain-nixpkgs` for everything to work smoothly: https://github.com/holochain/holochain-nixpkgs/pull/44.

overlays/holochain-nixpkgs/default.nix
```diff
@@ -3,8 +3,8 @@ let
   holochainNixpkgs = final.fetchFromGitHub {
     owner = "holochain";
     repo = "holochain-nixpkgs";
-    rev = "1e2067cecb4d7c02fc38ac407ec84f3c6e25a402";
-    sha256 = "sha256-CSEUkxMNC5VpKm75fJsYPE9Harbuy5m2xfXUOkkGqw0=";
+    rev = "8afbe20786130cf94a6f5d95293b5d6fa445cde6";
```
cole-h, line 6: This is the unmerged commit from https://github.com/holochain/holochain-nixpkgs/pull/44. The git revision will probably change once it's merged.

overlays/holochain-nixpkgs/default.nix
cole-h: Ah, alright. I've removed them in https://github.com/Holo-Host/holo-nixpkgs/commit/cc9c70ff997688efbafa2dac9443ab5ac61fec57.

grahamc:
This is blocked on https://github.com/holochain/holochain-nixpkgs/pull/44.

cole-h:
GitHub is weird when you delete the base branch of a PR and won't let you change the base branch and reopen it... New PR is up at https://github.com/Holo-Host/holo-nixpkgs/pull/1079.

PR #1065, opened 2022-01-18
host-console-ui: Proper signing

timotree3:
Uses https://github.com/Holo-Host/host-console-ui/pull/40

This PR shouldn't block release, so we can wait for host-console-ui to reviewed and merged into develop before merging this holo-nixpkgs PR


alastairong (approved):
Tested and approved

PR #1066, opened 2022-01-19
Fix `acme-*` cert service on match-server profile

peeech:
This PR is fixing an issue, where `acme-network-statistics.holo.host` service is unable to renew certificates from letsencrypt.com.


peeech:
calls from `acme-network-statistics.holo.host.service` to  `http://network-stat
istics.holo.host/.well-known/acme-challenge/` folder result in 402 not found.

After analyzing nginx configuration what is missing is a part that gives access on port 80 to folder `/var/lib/acme/acme-challenge`. There is a part that grans access on port 443 to this folder, but acme service is trying to make a http call, not https.

alastairong (approved):

PR #1072, opened 2022-01-21
hydra-master: update to c280692f

grahamc:
Resolves an issue with Hydra building locally, where Hydra would emit:

    possibly transient failure building ‘/nix/store/X.drv’ on ‘localhost’:
    dependency '/nix/store/Y' of '/nix/store/Y.drv' does not exist,
    and substitution is disabled


peeech (approved):
Tested on hydra-dev, ready to merge

PR #1077, opened 2022-01-25
Change Registration Flag Location

alastairong:
Moves `zt-auth-done-notification`  and `TLS-set-notification` flags from /var/lib/configure-holochain to /var/lib/holo-auth.

This is necessary so that we don't accidentally wipe these flag files while wiping keys as part of normal dev processes.


peeech (approved):
looks great!

alastairong:
Test that this works

PR #1085, opened 2022-01-31
HPOS reset also resets channel and configuration.nix

alastairong:


overlays/holo-nixpkgs/hpos-reset/hpos-reset.sh
```diff
@@ -0,0 +1,25 @@
...
+model=$(nixos-option system.hpos.target 2>/dev/null | sed -n '2 p' | tr -d \")
+
+if [[ -z "$model" ]]; then
+    model="holoport"
+fi
+
+echo "
+{
+  imports = [
+    <holo-nixpkgs/profiles/targets/$model>
```
timotree3, line 15: I'm worried that we will rename something in the `targets` folder, and this script will get out of date, and then once someone runs hpos-reset, we will not be able to SSH into their holoport because it is no longer configured as a holoport. Maybe that problem is not this script's fault. If we rename something in the `targets` folder, all the existing `configuration.nix` will get out-of-date anyway. Any thoughts?

overlays/holo-nixpkgs/hpos-reset/hpos-reset.sh
```diff
@@ -0,0 +1,25 @@
...
+    ./hardware-configuration.nix
+  ];
+}
+" > /etc/nixos/configuration.nix    
+
+nix-channel --add https://hydra.holo.host/channel/custom/holo-nixpkgs/master/holo-nixpkgs # We assume that the only channel aliases would be holo-nixpkgs and maybe nixos
+rm -rf /var
+reboot
+
+exit 0
```
timotree3, line 25: What's the point of this line of code? I would expect that the script exits with status `0` by default, plus I would expect that it never actually exists because `reboot` shuts down the machine before exiting.

overlays/holo-nixpkgs/hpos-reset/hpos-reset.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/sh
+# Simple shell script for resetting a HoloPort to default settings and clearing user
+##
+
```
timotree3, line 4: It would be good to configure the shell in the standard way for scripts.
```suggestion
set -ex
```

`set -e` means that the script will exit if any of the commands errors, rather than proceeding as normal. `set -x` means that the script will print out each command it's running as it runs them

overlays/holo-nixpkgs/hpos-reset/default.nix
```diff
@@ -0,0 +1,22 @@
...
+{
+
+  hpos-reset = stdenv.mkDerivation rec {
+    name = "hpos-reset";
+    src = gitignoreSource ./.;
+
+    nativeBuildInputs = [ makeWrapper ];
+
+    installPhase = ''
+      install -Dm 755 hpos-reset.sh $out/bin/${name}
```
timotree3, line 14: 755 means `rwxr-xr-x`, right? (User `root` can read, write, execute. Group `root` can read and execute. Other users can read and execute)

From the documentation of `install`, it looks like that's the default
```
  -m, --mode=MODE     set permission mode (as in chmod), instead of rwxr-xr-x
```
So why not just remove `-m 755`?
```suggestion
      install -D hpos-reset.sh $out/bin/${name}
```

overlays/holo-nixpkgs/hpos-reset/hpos-reset.sh
alastairong: Good point. I copied it over from another script and didn't think to remove it

overlays/holo-nixpkgs/hpos-reset/hpos-reset.sh
alastairong: Yea /targets/ is hardcoded locally and if we change the location it would cause issues. Most likely, if we were to change the location we would have a pointer file of some kind in /targets/ that points or imports the proper one from elsewhere?

overlays/holo-nixpkgs/hpos-reset/hpos-reset.sh
alastairong: I removed `set -e` because otherwise the script fails at `rm -rf /var` because for some reason it throws some weird error about being unable to delete `/var/empty`

overlays/holo-nixpkgs/hpos-reset/default.nix
alastairong: we have this in the update script too so I didn't want to tinker around with it. I can give it a go?

timotree3: It's fine to leave it. I'll resolve this comment
> I can give it a go?
Not sure I understand this question

overlays/holo-nixpkgs/hpos-reset/hpos-reset.sh
timotree3: In that case you can add `|| echo "ignoring failure from rm command"` after the `rm` command in specific

alastairong:
@peeech @zo-el Do you think this reset should also repair $NIX_PATHS? If so, any idea what the correct $NIX_PATHS are?

peeech (approved):
Great!

PR #1092, opened 2022-02-04
release.nix: fix channel constituents

cole-h:
Closes https://github.com/Holo-Host/bugs-n-issues/issues/28.

---

There were a couple of issues that were preventing the proper functionality of the release channel:

1. The `constitute` function was filtering out all items passed to it that were not a derivation. However, `attrNames` just returns a list of strings, so it would always create an empty list.
2. The `holo-nixpkgs-tests` attrset was just that: an attrset of derivations. Despite being tagged with `recurseIntoAttrs`, both the old and new constitute functions would not have thought `holo-nixpkgs-tests` was a derivation (because it wasn't). I decided to split the tests out of the overlay and move them directly into `release.nix`. This allows us to add the test derivations as constituents directly to the channel, rather than having to special-case `holo-nixpkgs-tests` and do some pre-processing on it.

---

~~**NOTE**: This PR is marked a draft because I want to break a test and see if Hydra catches it this time.~~ Both Hydras caught it :)

TODO:

- [x] Check that Hydra successfully builds this PR and creates a channel now
- [x] Break a test on purpose and check that Hydra fails to build it and doesn't create a channel
- [x] Unbreak it and mark this PR ready-for-review


peeech:
It would be great to change base of this PR to `hydra` branch.

cole-h:
Here's an eval showing the channel failed when the tests failed: https://hydra-dev.holo.host/eval/2206495 (on the dev Hydra) and https://hydra.holo.host/eval/1920850 (on the old Hydra).

I think this is ready, finally!

release.nix
```diff
@@ -28,12 +29,20 @@ in
...
-    self // {
+    self // tests // {
       holo-nixpkgs = releaseTools.channel {
         name = "holo-nixpkgs";
         src = gitignoreSource ./.;
-        constituents = constitute (attrNames self);
+        constituents = [ ]
+          ++ constitute (attrValues self)
+          ++ constitute [ tests ]
+        ;
```
peeech, line 46: This looks so obvious once you have made this change. I would never be able to do it myself though :-). Thank you!

PR #1093, opened 2022-02-07
nightly-test-runner: do not freak out on test failure

peeech:


modules/services/nightly-test-runner.nix
```diff
@@ -49,8 +49,8 @@ let
     function run_and_rename {
       local NEW_RESULT_FILE="OK-$RESULT_FILE"
 
-      run-nightly-test | tee $RESULT_FILE
-      if [[ $? != 0 ]]; then
+      local status=$(run-nightly-test > $RESULT_FILE 2>&1)$?
```
timotree3, line 52: This line is kind of weird. We're concatenating the output of `run-nightly-test > $RESULT_FILE 2>&1` with its status code. I guess we're just relying on its output being nothing, since we piped it into a file? Seems clearer to do something like this
```suggestion
      run-nightly-test > $RESULT_FILE 2>&1
      local status=$?
```

PR #1099, opened 2022-02-08
Infra 4 Fast Follow Release - Staging

alastairong:
- Better HPOS reset that resets config files and channel
- Patch for broken host console UI api calls

Will be QA'd and released together with https://github.com/Holo-Host/holo-auth/pull/55


peeech (approved):
Approved

PR #1079, opened 2022-01-26
overlays: init holochain-nixpkgs overlay

cole-h:
Supersedes https://github.com/Holo-Host/holo-nixpkgs/pull/1063.
Addresses https://github.com/Holo-Host/bugs-n-issues/issues/21.

---

Opening a new PR because GitHub is weird when you delete the base branch of a PR (I can't change the base to `develop` in https://github.com/Holo-Host/holo-nixpkgs/pull/1063...) -- sorry for the noise.

PR #1102, opened 2022-02-08
treewide: pkgconfig => pkg-config

peeech:


peeech:
For some strange reason Hydra is reporting an error [[link](https://hydra.holo.host/jobset/holo-nixpkgs/1102#tabs-errors)]:
```
error: undefined variable 'setuptools_scm' at [1m/nix/store/bwvnb2mwqrcz6bjgmd9j7gqpq2pxhk3g-source/overlays/holo-nixpkgs/default.nix[0m:328:60
```
`setuptools_scm` IS available in the scope of `python3Packages` and it is possible to build simple locally via `nix-build . -A simp_le`, so this error is obviously a bug.

cole-h:
Maybe that's because `setuptools_scm` is an alias to `setuptools-scm` -- if `simp_le` is used in the test suite and the error is bubbling up from there, that would be why: https://github.com/NixOS/nixpkgs/commit/3edde6562e19698da69a499881e0a2e4f5a497a2 made the tests disallow aliases.

peeech:
This was actually a part of a PR #1079

PR #1110, opened 2022-02-09
[holofix]: overlays/holo-nixpkgs/membrane-proof-service: set access to prod and dev urls

zo-el:


alastairong:
@zo-el Please point this into develop and #1094 instead. 

We have real users in pre-release who may still need to re-register.

zo-el:
K, pointed this to PR #1094 and this PR #1109 points to develop

PR #1113, opened 2022-02-10
release.nix: allow building unfree packages

cole-h:
The run-nightly-test package / module requires the google-chrome package
for puppeteer-based testing, but that is unfree. There are other
blockers preventing using other chromium-based browsers, so just allow
unfree packages.

---

Fixes https://github.com/Holo-Host/bugs-n-issues/issues/29.

---

~~I'm still rebuilding things (haven't gotten to the nightly-test yet), but it actually evaluated locally, so I'm taking that as a good sign.~~ `run-nightly-test{,as-ci}` succeeded!


peeech (approved):
Seems to be working for me too.

PR #1103, opened 2022-02-08
overlays/hydra: fix restrictEval patch

peeech:
Fixes patch that turns off restricted evaluation mode

PR #1115, opened 2022-02-11
overlays/holo-nixpkgs: override Hydra from upstream

cole-h:
The upstream package is provided under the attribute `hydra`, while the
Nixpkgs package is provided under the attribute `hydra-unstable`.


grahamc:
This is ultimately why the patches were appearing to not be applied to the new Hydra instance. I suspect this is fixing the underlying issue which #1103 was trying to fix.

peeech (approved):
:+1:

alastairong (approved):

PR #1117, opened 2022-02-14
Restrict build of run-nightly-test-* to x86_64 architecture

peeech:
because google-chrome (which run-nightly-test-* depend on) cannot be built on aarm machines


timotree3 (approved):
Would it be better to exclude arm from `platforms.linux`? (In case there's other Linux platforms that will work besides ARM and x86_64)

zo-el (approved):

peeech:
`platforms.linux` is a shorthand of `[ x86_64 aarm ]`. So in the realm on NixOs those two are equivalent. The other reason can be that `google-chrome` package [is explicitly restricting](https://github.com/NixOS/nixpkgs/blob/master/pkgs/applications/networking/browsers/google-chrome/default.nix#L169) builds to `"x86_64-linux"`

PR #1119, opened 2022-02-15
Bump to holochain 0.0.126 / HDK 0.0.122

alastairong:
- [x] Bump Holochain to 0.0.125
- [x] bump rust to v1.58.1
- [x] Servicelogger
- [x] HHA
- [ ] hpos-holochain-api (as needed)
- [ ] hosted-happs-monitor (as needed)
- [x] holo-auto-installer (as needed)
- [x] configure-holochain (as needed)
  - [x] handle read-only memproofs for other holonix-pkgs servers


alastairong:
@zo-el Configure-holochain and auto-installer are bumped for HDK 0.0.122. Do we need to bump everything else again too? Holochain is 0.0.125 and Servicelogger HDK 0.0.121.

PR #1127, opened 2022-02-23
Auto-update `holochain-nixpkgs` overlay

firestack:
Addresses Holo-Host/bugs-n-issues#31

---

This factors out `rev` and `sha256` into a json file which can be updated by `nix-prefetch-github` which has been configured using GitHub actions to run weekly and submit a PR if there are any changes.

in addition, this updates `holochain-nixpkgs` to the latest available commit and fixes the problems that arose, specifically `mkRust` not being available due to the usage of `final.callPackage` which provided a `pkgs` argument to the `default.nix` function effectively overwriting `pkgs.rust.mkRust`


cole-h:
Also cc @steveeJ

overlays/holochain-nixpkgs/default.nix
```diff
@@ -1,13 +1,15 @@
 final: prev:
 let
-  holochainNixpkgs = final.fetchFromGitHub {
+  holochainNixpkgs = let
```
steveeJ, line 3:
```suggestion
  holochainNixpkgsPath = let
```

overlays/holochain-nixpkgs/default.nix
```diff
@@ -1,13 +1,15 @@
...
+  in final.fetchFromGitHub {
     owner = "holochain";
     repo = "holochain-nixpkgs";
-    rev = "37e8512602030c4601742fdc71a66bf6e8b966a4";
-    sha256 = "k2bbSjkQOYgcGWbEEBR8A8u7SV/4YC3lneqKWdu9ZII=";
+    inherit (srcDef) rev sha256;
   };
 
-  holochainMeta = final.callPackage "${holochainNixpkgs}/packages/holochain" { };
+  holochain = (import holochainNixpkgs { });
```
steveeJ, line 11:
```suggestion
  holochainNixpkgs = (import holochainNixpkgs { });
```

overlays/holochain-nixpkgs/default.nix
```diff
@@ -1,13 +1,15 @@
...
+  holochainMeta = final.callPackage "${holochainNixpkgs}/packages/holochain" { mkRust = holochain.pkgs.rust.mkRust; };
```
steveeJ, line 12:
```suggestion
  holochainMeta = final.callPackage "${holochainNixpkgs}/packages/holochain" { mkRust = holochainNixpkgs.pkgs.rust.mkRust; };
```

steveeJ (commented):
i have a different understanding of the named values so i made some naming suggestions.

also you may want to rerun the `update-holochain-versions` tool on each update run to ensure the generated version file is up to date. you can hardcode the rust version into the call of that binary.

steveeJ:
@zo-el ran into issues in https://github.com/Holo-Host/holo-nixpkgs/pull/1135 that suggest there's a bug in the *holochain-nixpkgs*'s _holochain.nix_ function. it seemed that the `rustVersion` attribute in the `hpos.nix` file is not consistently used. however if that is the case then i would expect for the build to fail in this PR too, which doesn't seem the case. 

before merging this PR i suggest to double-check that the version specified in `hpos.nix` is actually built.

overlays/holochain-nixpkgs/holochain/versions/hpos.nix
```diff
@@ -27,9 +31,13 @@
...
         ];
 
+        rustVersion = "1.58.1";
+
         cargoLock = {
             outputHashes = {
             };
         };
+
+        rustVersion = "1.58.1";
```
zo-el, line 41: This is a good example of our CI's not working correctly. This duplication of `rustVersion` should cause the CI to fail because it cannot build on a holoport.

overlays/holochain-nixpkgs/holochain/versions/hpos.nix
```diff
@@ -12,12 +12,16 @@
...
 
+    rustVersion = "1.58.1";
+
     binsFilter = [
         "holochain"
         "hc"
         "kitsune-p2p-proxy"
     ];
 
+    rustVersion = "1.58.1";
```
zo-el, line 23:
```suggestion
```

overlays/holochain-nixpkgs/holochain/versions/hpos.nix
zo-el, line 41:
```suggestion
```

zo-el (changes requested):

overlays/holochain-nixpkgs/holochain/versions/hpos.nix
steveeJ: to support this point, running `nix-shell --run hpos-shell` for this pull-request fails with the following error

```
$ nix-shell --run "hpos-shell"
error: attribute 'rustVersion' already defined at /home/steveej/src/holo/holo-nixpkgs/overlays/holochain-nixpkgs/holochain/versions/hpos.nix:15:5

       at /home/steveej/src/holo/holo-nixpkgs/overlays/holochain-nixpkgs/holochain/versions/hpos.nix:23:5:

           22|
           23|     rustVersion = "1.58.1";
             |     ^
           24|
(use '--show-trace' to show detailed location information)
```

i agree that CI should be reliable in detecting such errors.

PR #1132, opened 2022-02-24
Hpos reset hotfix prerelease

alastairong:
Fixes bug with nixos-generate-config path
Stops nixos-generate-config from adding USB drive to filesystem causing bootup issues


overlays/holo-nixpkgs/hpos-reset/hpos-reset.sh
```diff
@@ -20,7 +20,7 @@ echo "
 }
 " > /etc/nixos/configuration.nix    
 
-nixos-generate-config
+nixos-generate-config --no-filesystems # So we don't declare the USB drive to be an integral part of our filesystem
```
peeech, line 23: I am not sure about it, this line means that there's gonna be no entry `fileSystems."/" = { device = "/dev/vda1"; fsType = "ext4"; };` in hardware-configuration.nix file.

And my question here is, why do we need to re-generate hardware-configuration at all? If we re-write configuration.nix it is already a lot of cleanup.

peeech (approved):
I think this should work well

PR #1133, opened 2022-02-24
Hpos reset hotfix staging

alastairong:
Fixes bug with nixos-generate-config path
Stops nixos-generate-config from adding USB drive to filesystem causing bootup issues


peeech (approved):
I think this should work well

PR #1134, opened 2022-02-24
Nightly tests uses dummy hApp declared in PR

alastairong:
Dummy hApp URL and DNAs are declared in global-settings.nix
If either is missing nightly tests will fail

Companion to https://github.com/Holo-Host/nightly-network-test/pull/21


peeech (approved):
Let's see how it works :-)

PR #1135, opened 2022-02-28
Bump holochain to v0.0.127

zo-el:
- [x] Bump Holochain to 0.0.127
- [x] bump rust to v1.58.1
- [x] Servicelogger
- [x] HHA
- [x] EC
- [x] Joining-code-factory
- [x] holo-auto-installer (as needed)
- [x] configure-holochain (as needed)


zo-el:
@peeech this PR does not build on a holoport. It passes all the hpos tests, but fails to build on any holoport. 
It looks like holochain is not able to build. 
Can you take a look

peeech:
Yes in fact I am seeing an issue during build of a system:
```
trying https://github.com/holochain/holochain-nixpkgs/archive/7812b7a92fbb02e1950de9893dd7a0ba1a551494.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   169  100   169    0     0    684      0 --:--:-- --:--:-- --:--:--   686
100 9250k    0 9250k    0     0  4606k      0 --:--:--  0:00:02 --:--:-- 8025k
unpacking source archive /build/7812b7a92fbb02e1950de9893dd7a0ba1a551494.tar.gz
unpacking 'https://github.com/nixos/nixpkgs/archive/ed02c2ba0384b2800db41333045a6fb781f12aac.tar.gz'...
error: attempt to call something which is not a function but a list, at /nix/store/vnfjwcv41b3mrd79iaba8592lxp9p1pm-source/lib/fixed-points.nix:69:67
```
This is not a compilation error, but a configuration evaluation error, which is pointing to the fact that there's an error in an underlying nix package/code.

peeech:
It definitely requires deeper research to identify where is root cause of the problem. I am thinking the best thing to do is move this to an issue on `holochain-nixpkgs` repo. Has there been any conversation or attempt to fix it before @zo-el ?

zo-el:
We attempted to build holochain which was failing locally too, which is why we changed the build process of holochain here.

peeech:
Ok, I am not sure what are the next steps on this one. Is there a conversation happening somewhere where it would be discussed?

alastairong:
I think stefan and joel are pairing on this @peeech

overlays/holochain-nixpkgs/default.nix
```diff
@@ -1,21 +1,25 @@
...
     repo = "holochain-nixpkgs";
-    rev = "37e8512602030c4601742fdc71a66bf6e8b966a4";
-    sha256 = "k2bbSjkQOYgcGWbEEBR8A8u7SV/4YC3lneqKWdu9ZII=";
+    inherit (srcDef) rev sha256;
   };
 
-  holochainMeta = final.callPackage "${holochainNixpkgs}/packages/holochain" { };
+  holochainMeta = final.callPackage "${holochainNixpkgsPath}/packages/holochain" {
+    # FIXME: this will cause the rustVersion of the hpos.nix file to be ignored.
+    mkRust = { ... }: final.rust.packages.stable.rust.rustc;
```
steveeJ, line 13: @peeech i'd like to discuss this hack and potentially find a better solution which could involve refactoring the holochain expression.

the function that's expected here is used to instantiate [a rust toolchain given a track and a version][(https://github.com/holochain/holochain-nixpkgs/blob/1f01f82d23a0bff9c7f1d4b11b3c7711808aaceb/packages/holochain/default.nix#L26](https://github.com/holochain/holochain-nixpkgs/blob/1f01f82d23a0bff9c7f1d4b11b3c7711808aaceb/overlays/rust.nix#L17)). this is from the holochain-nixpkgs repo itself:

```nix
defaultExtensions = [
        "rust-src"
      ];

      defaultTargets = [
        "aarch64-unknown-linux-musl"
        "wasm32-unknown-unknown"
        "x86_64-pc-windows-gnu"
        "x86_64-unknown-linux-musl"
        "x86_64-apple-darwin"
      ];

      mkRust = { track, version }: self.rust-bin."${track}"."${version}".default.override {
        extensions = self.rust.defaultExtensions;
        targets = self.rust.defaultTargets;
      };
```

you see that it uses parts of [oxalica/rust-overlay](https://github.com/oxalica/rust-overlay/)https://github.com/oxalica/rust-overlay/) which is an alternative to mozilla-nixpkgs.

hence, we need to define a `mkRust` that is compatible with your repository and preferences and pass that in. alternatively we could find a way that makes it acceptable that holo-nixpkgs uses the one that's in holochain-nixpkgs.

yet another alternative is that we keep my hack, which means that you'll always use the stable rust from this repo and the holochain expressions cannot instantiate a rust toolchain based on the given `rustVersion`.

/cc @firestack

zo-el: Let's continue this conversation in Issue #1172 so that we can get this PR reviewed and merged.

overlays/holo-nixpkgs/configure-holochain/default.nix
```diff
@@ -7,11 +7,11 @@ rustPlatform.buildRustPackage {
   src = fetchFromGitHub {
     owner = "Holo-Host";
     repo = "hpos-configure-holochain";
-    rev = "5c1510a08dc814fcc36a2999faa4cecca230b250";
-    sha256 = "0sypnrvi1jv4ap0sgqzimy7rqsy19n5xip4pfhkdgaij438f17p9";
+    rev = "d905df56e137111816e49bc46b9891eaa544e271";
```
timotree3, line 10: By our policy, this should include a comment saying why that version, right?

overlays/holo-nixpkgs/holo-auto-installer/default.nix
```diff
@@ -7,11 +7,11 @@ rustPlatform.buildRustPackage {
   src = fetchFromGitHub {
     owner = "Holo-Host";
     repo = "holo-auto-installer";
-    rev = "92b5c9ca91452ac67a0cfd9dde258175550d53d1";
-    sha256 = "0nvlj8rfah63mdjlsf2lxlmaalq7axmd9xr20rx54qpr2d744szw";
+    rev = "b3beb6e2350e439f1f7df6afd8a0a8e1f6c3320b";
```
timotree3, line 10: Same as above

profiles/logical/global-settings.nix
```diff
@@ -1,9 +1,9 @@
 {
   holoNetwork = "devNet";
-  DEV_UID_OVERRIDE = "1e0b99a2-f828-41a5-b410-33e81ee6f422";
+  DEV_UID_OVERRIDE = "develop-f828-41a5-b410-33e81ee6f422";
```
timotree3, line 3: This seems sketchy. Why not generate a new UUID?

timotree3 (approved):
LGTM. I haven't checked all the commit hashes we're bumping to, but I assume some test will fail if they're wrong

PR #1139, opened 2022-03-02
fix(hydra-configuration): Increase GC size to 60gb

firestack:
For the following formula, increase the maximum number of bytes freed by nix-collect-garbage from 15 (gb) to 60 (gb).  
`60(gb) - (space remaining on /nix/store/ disk in bytes)`

---
Addresses: https://github.com/Holo-Host/bugs-n-issues/issues/22/


peeech (approved):
Simple and works. Great!

PR #1140, opened 2022-03-02
feat(logical/holo): SSH without root (1/2)

firestack:
## Changes
This PR does the following
- Create "normal" user accounts for each SSH key that exists in `users.users.root.openssh.authorizedKeys.keys`.
- For each user
  - Set  `${username}.openssh.authorizedKeys` to contain the matching key in `root.openssh.authorizedKeys`.
  - Add user to `wheel` group to provide `sudo` access.
- Set `security.sudo.wheelNeedsPassword` to `false`

## Scope:
NixOS Configurations which import the nixos module `profiles/logical/holo/default.nix` will be affected. This is effectively the set of profiles under `profiles/logical/holo/`.

## User Experience:
It is expected that, when this change is deployed, users with current access via `root` are able to login using a username and SSH key already in their possession. After logging in, the user is expected to be able to execute commands using `sudo`, without providing a password.

It is also expected that the users will have no change to the existing workflow of logging in using `root` ensuring that any breaking changes introduced by PR can be reverted and addressed.

##  Next Steps:
To ensure the minimal chance of introducing irreversible breaking changes and losing SSH access to machines affected, this PR does half of the necessary work to complete https://github.com/Holo-Host/bugs-n-issues/issues/11.
The remaining work to be done is to remove SSH keys from the `root` user and is contained at this branch [DeterminateSystems/holo-nixpkgs#kf/11/remove-ssh-keys](https://github.com/DeterminateSystems/holo-nixpkgs/blob/kf/11/remove-ssh-keys/profiles/logical/holo/default.nix) to be merged after this PR is tested and confirmed to work as expected.

---
Addresses: https://github.com/Holo-Host/bugs-n-issues/issues/11

PR #1146, opened 2022-03-03
Add a key for DetSys

grahamc:
This key will let @firestack and @cole-h access the infrastructure, but in a way where their access is limited by DetSys's policies, and an internal audit log is created.

After validating this new key, we should delete my personal key.

Note: I'd prefer to apply this on top of @firestack's PR #1140, but if that is going to take a bit to apply we can do this one first and apply this to #1140 after.


alastairong (approved):

peeech:
Latest changes in #1140 make this PR obsolete.

grahamc:
Looks good, thanks @peeech!

PR #1147, opened 2022-03-04
Netstats deploy master

alastairong:


timotree3 (approved):
Blindly approving

zo-el (approved):

modules/default.nix
```diff
@@ -32,9 +32,10 @@ with lib;
     ./services/membrane-proof-service.nix
     ./services/kv-uploader.nix
     ./services/lair-keystore.nix
-    ./services/match-service-api.nix
+    ./services/hp-stats-api.nix
+    ./services/mock-hpos-config.nix
```
zo-el, line 36: This looks like some residue of some old code

alastairong: They look correct to me?

zo-el: `mock-hpos-config` is not a service, this is from the old code that was supposed to help with the hpos-tests

`./services/mock-hpos-config.nix`

zo-el: Its also in the other two PR's that you have merged

Issue #1081, opened 2022-01-27
NixOs tests are currently failing on develop


peeech:
Both of our NixOs tests located in [/tests](https://github.com/Holo-Host/holo-nixpkgs/tree/develop/tests) folder are failing on develop branch.

Current tip of develop is at https://github.com/Holo-Host/holo-nixpkgs/commit/71d62aa309ae5379a72eb4198e61cf676005aad3

peeech:
After running tests locally in a interactive mode I can see that tests are failing because VM is never able to boot up past `testScript.start_all()`.

I am seeing in the logs:
```
machine # [  274.411895] lair-keystore-pre-start[1220]: Lair Keystore Prestarting...
machine # [  274.496144] lair-keystore-pre-start[1220]: initializing keystore from config at /etc/hpos/config.json with password pass
machine # [  284.654008] lair-keystore-pre-start[1229]: Error: failed to open file /etc/hpos/config.json
machine # [  284.680196] lair-keystore-pre-start[1229]: Caused by:
machine # [  284.697170] lair-keystore-pre-start[1229]:     No such file or directory (os error 2)
```

peeech:
This seems to be caused by the fact that `/tests/hpos-api-tests` is requiring first starting VM in [line 17](hpos-api-tests), then creates `/etc/hpos/config.json` file in [line 22](https://github.com/Holo-Host/holo-nixpkgs/blob/develop/tests/hpos-api-tests/default.nix#L22). But VM at boot time is trying to start [`lair-keystore.service`](https://github.com/Holo-Host/holo-nixpkgs/blob/develop/profiles/logical/hpos/sandbox/default.nix#L96), which in turn is trying to access not-yet-created `/etc/hpos/config.json` and fails, VM is never able to start.

alastairong:
Strange that these tests passed on their respective PRs...

peeech:
It does not seem possible to create a file before VM is up and running. One solution I see can be to create a service which runs before `lair-keystore.service` that creates `/etc/hpos/config.json` in the same manner as [line 22](https://github.com/Holo-Host/holo-nixpkgs/blob/develop/tests/hpos-api-tests/default.nix#L22). This service would be only active on a [sandbox profile](https://github.com/Holo-Host/holo-nixpkgs/blob/develop/profiles/logical/hpos/sandbox/default.nix).

zo-el:
[...]

PR #1161, opened 2022-03-07
Fix broken tests

alastairong:


zeeshan595 (approved):

profiles/logical/hpos/sandbox/default.nix
```diff
@@ -95,6 +95,8 @@ in
 
   services.lair-keystore.enable = true;
 
+  services.mock-hpos-config.enable = true;
```
zo-el, line 98: @peeech 
Did you get a chance to review PR #1091. And if we are going ahead with the mock-hpos-config approach let's close out Issue #1081

SirRobLyon (approved):

peeech:
fixed in #1161

PR #1162, opened 2022-03-07
 feat(logical/holo): SSH without root (2/2)

firestack:
Now that https://github.com/Holo-Host/holo-nixpkgs/pull/1140 has been merged and [confirmed to be working](https://github.com/Holo-Host/bugs-n-issues/issues/11#issuecomment-1060714184), we can move ahead remove the previously used user keys from the `root` user account.

## Changes
This PR does the following
- Remove user keys from `root.openssh.authorizedKeys`
- Finalize removing user `grahamc` 

## Scope:
NixOS Configurations which import the nixos module `profiles/logical/holo/default.nix` will be affected. This is effectively the set of profiles under `profiles/logical/holo/`.

## User Experience:
It is expected that, when this change is deployed, users with current access via `root` are no longer able to login to machines using `root`.

---
Addresses: https://github.com/Holo-Host/bugs-n-issues/issues/11


peeech (approved):
:+1:

PR #1175, opened 2022-03-10
overlays/holo-nixpkgs/aorura:update: only for linux

zo-el:


peeech:
It is hard for me to review because I am not sure what issue is this trying to address. Can you describe?

Also pushing fix directly (and only) to master does not seem to me like a good solution. All other branches will stay affected with the same issue.

zo-el:
> It is hard for me to review because I am not sure what issue is this trying to address. Can you describe?

This is a fix for the failing build in hydra: https://hydra.holo.host/jobset/holo-nixpkgs/master
 
> Also pushing fix directly (and only) to master does not seem to me like a good solution. All other branches will stay affected with the same issue.

I completely agree, and I would never do it this way. But it looks like you guys don't have a pattern for merging into master, staging and pre-release. All those branches have different git histories and commits have been manually cheery picked and merged in.
Once all those branches are back to the same history, I am happy to follow the pattern that is proposed.
Currently, I just made this PR to see what you think of it. if you approve I can merge it into the other branches too.

alastairong (dismissed):

alastairong:
Thanks @joel. Totally agree there's no need for this for x64-darwin.

Rather than create companion PRs for `pre-release` and `staging`, I would like to hold off for now and maybe totally replace staging/pre-release with the NixOS bump.

I am getting quite fed up of this master/staging/pre-release system and I think very soon is the time to replace it.

peeech:
@zo-el @alastairong Do we want to make this PR into develop instead?

alastairong:
It's not necessary since the issue doesn't happen on develop, but maybe we should anyway since it's silly to build this package for darwin.

Sooooo...Yes let's do that!

alastairong:
I will shortly close this and create one for develop

alastairong:
Closing in favour of #1185

PR #1202, opened 2022-03-28
Bump holochain v0.0.131

zo-el:


modules/services/holochain-proxy.nix
```diff
@@ -23,7 +23,7 @@ in
     };
 
     package = mkOption {
-      default = pkgs.kitsune-p2p-proxy;
+      default = pkgs.kitsune-p2p-tx2-proxy;
```
alastairong, line 26: @peeech Do you know if this is the right one to use for the proxy service?

alastairong (approved):

PR #1208, opened 2022-03-31
profiles/logical: change network names

alastairong:
With new alpha channel, change network names so that alphaNet is for alpha channel and mainNet is for master


timotree3 (approved):
According to the hackmd, we should have four branches
```
master mainNet
alpha alphaNet (old staging)
staging (old prerelease)
develop devNet
```

Is this going in that direction? What `Net` do our branches currently specify?

I guess I'll just approve and trust you

alastairong:
> According to the hackmd, we should have four branches
> 
> ```
> master mainNet
> alpha alphaNet (old staging)
> staging (old prerelease)
> develop devNet
> ```
> 
> Is this going in that direction? What `Net` do our branches currently specify?
> 
> I guess I'll just approve and trust you

Thanks for asking. Yes this, plus #1207 are going in that direction. We have 4 auto-built channels (master, alpha, staging, develop). We currently have 3 nets though:
```
mainNet -> master
alphaNet -> alpha
devNet -> staging
devNet -> develop
```

Down the line I imagine it will be good to have a separate staging vs develop, but for where we are in maturity it doesn't seem worth it.

If we have stageNet and devNet with different components, then it adds cost and maintenance by duplicating components that are already barely used.

If we have declare the 2 nets but have the same components across both, then it adds a small but IMO unnecessary bit of work to ensure everything is on the right channel every time we deploy.

PR #1212, opened 2022-04-01
131 for staging

alastairong:


profiles/logical/global-settings.nix
```diff
@@ -1,9 +1,9 @@
 {
   holoNetwork = "devNet";
```
zo-el, line 2: Does staging use devNet or alphaNet?

zo-el (approved):
Just left one question

profiles/logical/global-settings.nix
alastairong: It shares devNet with develop but different UID. alpha uses alphaNet and master uses mainNet

PR #1214, opened 2022-04-05
Fix holochain-api permissions

alastairong:
holochain-api gets the following issue on `/register-happ`

`hpos-holochain-api error: Error: Error: EACCES: permission denied, open '/var/lib/configure-holochain/config.yaml'`


modules/services/hpos-holochain-api.nix
```diff
@@ -25,20 +25,15 @@ in
...
+        User = "configure-holochain";
+        Group = "configure-holochain";
         UMask = "0002";
         Restart = "always";
       };
     };
     
     systemd.tmpfiles.rules = [
-      "d /run/hpos-holochain-api 0770 hc-api apis - -"
+      "d /run/hpos-holochain-api 0777 configure-holochain apis - -"
```
alastairong, line 36: Root temporarily needs access because we are publishing apps using the `hpos-holochain-api` CLI.

PR #1216, opened 2022-04-05
[TEST] reproduce lair-prestart error

peeech:
In this PR I am trying to reproduce test errors discussed in #1081  and fixed by #1161. An intention here is to revert #1161 and see if #1091 (which is much simpler than #1161) would fix an error as well.
- [x] revert #1161 and reproduce error
- [x] apply #1091 and check result
- [x] act accordingly to an outcome :-)


peeech:
Tests revealed flaw in systemd service order where netstatsd needs to depend on `configure-holochain.service` rather than `holochain.service`

PR #1218, opened 2022-04-05
[hotfix]: mem-proof path dynamically updated based on net and uid

zo-el:
- [x] dynamic mem-proof file paths, based on network and uid
- [x] configure-holochain uses MEM_PROOF_PATH as globally provided
- [x] holo-auth calls mem-proof server when mem-proof is not found and ZT-registration based on zt-status
- [x] dev-mem-proof-server holoAgentOveride updated


alastairong (approved):
Have confirmed this works great!

peeech (approved):
I have not reviewed `configure-holochain` nor `holo-auth` updates to the code.  Holo-nixpkgs code looks good.

PR #1220, opened 2022-04-06
Add promtail log push to grafana to holo Network components

peeech:
In our efforts to upgrade security we start storing server logs externally. In this PR we are adding promtail service that pushes to Grafana cloud logs from the following machines:
- [x] `holochain-proxy` profile
- [x] `match-server` profile
- [x] `membrane-proof-server` profile
- [x] `router-gateway` profile


alastairong (approved):
Looks straightforward

PR #1223, opened 2022-04-07
Add HPOS_CONFIG_PATH to netstatsd env

peeech:
fixes error where `HPOS_CONFIG_PATH` is not present in `netstatsd.service` environment:
```
Apr 07 10:36:12 hpos netstatsd[748648]: Error: Cannot read HPOS_CONFIG_PATH from env var
```


peeech:
I have confirmed it is fixing an issue

zeeshan595 (approved):

PR #1228, opened 2022-04-14
Add detsys ssh key to HPOS

peeech:
In this PR an ssh key for determinate.systems authority is added. This allows detsys team access Holoports with `profile.development` activated (which is ON by default on all the holoports).


marycamacho (approved):
What will the process for removing access be?

peeech:
> What will the process for removing access be?

Deleting line added in this PR will revoke access. @marycamacho

PR #1232, opened 2022-04-20
Bump hash space and new environment path

robbiecarlton:


timotree3 (approved):
We've received green light in MM

PR #1235, opened 2022-04-22
feat(physical/hpos/holoport): add fan control configuration profile 

firestack:
After investigation and experimentation with the Holoport, we've designed a NixOS module which configures a Holoport to use a user controllable fan curve, from within NixOS using the `fancontrol` script.

This PR adds a default fan control configuration which has similar characteristics to the fan curve defined in the BIOS by default, but is instead controlled by `fancontrol` on a 10 second interval.

When a Holoport reboots, the `sysfs` path `/sys/class/hwmon` is populated with `hwmon` devices non deterministically. Because `fancontrol` requires that it's devices are referenced either via absolute path, or the non deterministic `hwmon` id, We instead use `udev` rules to match the devices we're looking for and create named links we can reference in `/dev/holo` to the devices necessary for `fancontrol`.

Finally, the `Holoport` requires the `it87` kernel driver to control the cpu fan plugged into the `Holoport`.

I want to note about the placement of this module within `profiles/physical/hpos/holoport`. This module is specifically compatible with the `Holoport` and `Holoport-plus`, and it is not currently known what would be needed to make this module usable by by the `Holoport-nano`.

We monitored the audio from the fans and they remained quiet and temperatures low while running `stress` for several minutes.

---

Related to: https://github.com/Holo-Host/bugs-n-issues/issues/3


alastairong:
@firestack 

Per your request for feedback, in theory I think `fancontrol.nix` file should be hardware-specific. So one file in ./hpos/holoport` and one in `./hpos/holoport-plus`, since these have different hardware and may require different thermal management approaches.

But since HP and HP+ are so similar in practice, and the HP Nano doesn't exist, perhaps we can just have it in the parent `./hpos` directory

@peeech Thoughts?

peeech:
The location of the file is fine. `fancontrol.nix` is `physical/hpos` specific and profiles import it as needed.

I checked documentation of [pwmconfig](https://linux.die.net/man/8/pwmconfig)(8). I am curious how did you adjust values in `hardware.fancontrol.config`?

peeech:
@firestack This looks really good, thanks for all the documentation. I will run tests on my Holoport over the weekend.

firestack:
Hey @peeech @alastairong !

I reworked it a little bit to make it possible to configure the HP and HP+ separately from each other while sharing the module in a reasonable location. Let me know if you want me to move the files around at all, or remove redundancy[*1]. I haven't been able to hear the HP+ first hand yet but I can tweak the HP+ config numbers to get a better sound ramp if needed.

I'll be curious how your test goes @peeech!


[1]: duplicated the config into the `hpos/holoport/default.nix` file to make it clear how it's configured and show the symmetry between the hp and hp+ files.

peeech (approved):
Tests are showing no peaks in fan speed. Well done!

peeech:
To answer your question @firestack I switched my Holoport to this PR. As expected at the time of boot I was able to hear a peak in fan activity, but then (once config kicks in I guess) when I was compiling Rust locally I could see an improvement in fan's activity - no peaks during build execution

PR #1249, opened 2022-05-04
[rsm]: auto-installed does not instal read-only when special installed app id is found

zo-el:


overlays/holo-nixpkgs/hpos-holochain-api/src/api.js
```diff
@@ -9,13 +9,15 @@ const util = require('util')
 const installHostedHapp = async (
   happId,
   bundleUrl,
+  isClone,
+  special_installed_app_id,
```
pegaltier, line 13: ugly to mix different case formatting (camel case vs snake case)

PR #1257, opened 2022-05-10
fix(logical/hpos/acme-.service): avoid infinite loop when `$HPOS_CONFIG_PATH` is missing

firestack:
if `hpos-config-into-base36` fails to read $HPOS_CONFIG_PATH then the
script will exit and an empty stdout will be written to the `base36_id`
variable.
Curl will then use this empty variable to query a URL that
will not have the expected information, forever.

This PR implements an early exit by testing if the variable is empty and exiting 1 if it is.
This fixes the infinite loop by preventing the $base36_id variable from being empty and curl failing in a loop. 
This also adds a 1h (3600s) timeout on the service so that we cannot sit in a stuck state forever.

---

Addresses https://github.com/Holo-Host/bugs-n-issues/issues/44


peeech (approved):
:+1:

PR #1262, opened 2022-05-17
role must match registration code role

zeeshan595:
* Removed role from pubKey object in mongo db
* Verifies the agent pub key is the same role as the registration code it is being inserted in


overlays/holo-nixpkgs/membrane-proof-service/src/helper/register-user.js
```diff
@@ -56,8 +62,7 @@ async function registerUser(res, payload) {
 
     // Update ops-console db after Successful Verification
     codeList.agentPubKeys.push({
-        role,
```
JettTech, line 59: Why remove this role? Is it because it seems redundant with the `RegistrationCode.role` above? I think we were keeping this here until agent holo hashes included a role prefix and are expecting it to be there in a couple places until then.

JettTech: That said, I'm ok with removing it from the agent key list as long as we make the schema change known across teams to avoid and unintended consequences.

JettTech (approved):
LTGM.  I left one small Q with a potential TODO for cross team communication.  Otherwise, this should cover it. :+1:

overlays/holo-nixpkgs/membrane-proof-service/tests/End-To-End.json
```diff
@@ -0,0 +1 @@
...nsomnia.send();\nconst body = JSON.parse(response.data);\n\nexpect(response.status).to.equal(500);\nexpect(body.error).to.equal('Invalid registration code');\nexpect(body.isDisplayedToUser).to.equal(true);\nexpect(body.info).to.equal('No record found for not.found@email.com');","_type":"unit_test"},{"_id":"ut_34a96b9dc5f2414c83fa395bacf43179","parentId":"uts_0e307c34117845449d33059a4b4c3593","modified":1652977672997,"created":1652963879770,"requestId":"req_fee89890b1b3467bba98bb4a23907446","name":"Register User - Invalid Role","code":"const response = await insomnia.send();\nconst body = JSON.parse(response.data);\n\nexpect(response.status).to.equal(400);\nexpect(body.message).to.equal('the role provided does not match the registration code');","_type":"unit_test"},{"_id":"ut_ba767f016e4448f98f85b8bb67fec0a1","parentId":"uts_0e307c34117845449d33059a4b4c3593","modified":1652977674047,"created":1652972065374,"requestId":"req_7d7b639e11694cf8a96827935fb92968","name":"Register User - Success","code":"const response = await insomnia.send();\nconst body = JSON.parse(response.data);\nexpect(response.status).to.equal(200);\nexpect(body.mem_proof).to.be.an('string');","_type":"unit_test"}]}
```
JettTech, line 1: Can you add test instructions on how into the readme file in membrane-proof-service directory?  I this will be helpful given the repo breadth is so vast.

PR #1273, opened 2022-06-09
Handle duplicate domains from HHA

alastairong:
It's possible to register the same domain to multiple happ_ids
Since domains are keys in CF's KV store, this kills the environment as cannot be populated due to duplicate keys
This is a stop-gap solution to discard all duplicate domain records so that only the first happ gets registered.

Users will be able to publish a new hApp to the same domain as long as they delete the old one.

In future, we should handle this via HHA validation that returns an error to the UI for better UX.


overlays/holo-nixpkgs/hosted-happ-monitor/src/index.js
```diff
@@ -49,15 +49,17 @@ const main = async () => {
     id: happ.id
   }))
   
-  if (happList.length === 0) {
+  const uniquehAppList = happList.filter((urls => ({ url }) => !urls.has(url) && urls.add(url))(new Set));
```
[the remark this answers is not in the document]
zo-el: It does not return deleted entries. I have tests in HHA that cover that case.

overlays/holo-nixpkgs/hosted-happ-monitor/src/index.js
```diff
@@ -49,15 +49,17 @@ const main = async () => {
...
+
+  if (uniquehAppList.length === 0) {
     console.log("no happs published")
     return
   }
```
zo-el, line 57: This return will let the KV store stay as is when no apps are hosted. The KV store should be updated to the empty list. So that we stop hosting whatever was paused or stopped.

zo-el (changes requested):
one requested change

overlays/holo-nixpkgs/hosted-happ-monitor/src/index.js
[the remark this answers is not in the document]
robbiecarlton: Although funnily enough, I just posted an issue that `get_my_happs` *does* return them.

overlays/holo-nixpkgs/hosted-happ-monitor/src/index.js
alastairong: Good point! Updated - I think insertMany() works with an empty array but I haven't tested it.

alastairong:
Also taking this opportunity to bump hashspace because of too many dummy happs installed

overlays/holo-nixpkgs/hosted-happ-monitor/src/index.js
```diff
@@ -51,14 +51,14 @@ const main = async () => {
   
   const uniquehAppList = happList.filter((urls => ({ url }) => !urls.has(url) && urls.add(url))(new Set));
 
+  await upload(statsDb, uniquehAppList, 'happ_list')
+  console.log('happ list updated')
+
   if (uniquehAppList.length === 0) {
     console.log("no happs published")
     return
```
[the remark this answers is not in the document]
alastairong: I thought about that. I don't think we need to clear the happ2host KV store if we're not using it but I can change it if you feel strongly.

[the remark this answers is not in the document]
alastairong: @zo-el ping!

alastairong:
@zo-el Please review my update. It's not as simple as removing the return. What we really are trying to do is delete every `host_for_<happ_id>` and `<happ_id>_tranches` collection that corresponds to a happ_id no longer in the host list. So I've created a new function to do that.

alastairong:
@zo-el ping for review please

overlays/holo-nixpkgs/hosted-happ-monitor/src/index.js
```diff
@@ -95,6 +100,17 @@ const upload = async(db, data, collection_name) => {
   await collection.insertMany(data)
 }
 
+const clearOldTranches = async(db, happList) => {
+  let collection_list = await db.listCollections().toArray();
+  
+  collection_list.forEach(async (collection_name) => {
+    if (happList.some(happ_id => !collection_name.includes(happ_id))) {
```
zo-el, line 107: @alastairong I am not sure about this condition statement.

if anyone of the happ_id is not a part of the collection_name then drop the collection?

That should cause every collection to be dropped every time right?

alastairong: Oh whoops!
```suggestion
    if (happList.some(happ_id => collection_name.includes(happ_id))) {
```

zo-el: Even that condition will just delete/clean-up the collections that need to be updated. The Old collections(i.e. the happs that were stopped) will still remain.

Is that ok?

alastairong:
```suggestion
    if (happList.every(happ_id => !collection_name.includes(happ_id))) {
```

alastairong: @zo-el Definitely wasn't ok. Head was screwed on wrong

alastairong:
```suggestion
    if !happList.find( h => collection_name.includes(h)) {
```

PR #1279, opened 2022-06-21
[QA TESTING] - Publisher Portal .1 and .2

zo-el:


alastairong:
@zo-el if you're using this for testing reminder that:
1. You need to switch the match-server over if it's a new hash space
2. You need to include the code for #1273 as that is not yet in develop

zo-el:
> @zo-el if you're using this for testing reminder that:
> 
>     1. You need to switch the match-server over if it's a new hash space
> 
>     2. You need to include the code for [Handle duplicate domains from HHA #1273](https://github.com/Holo-Host/holo-nixpkgs/pull/1273) as that is not yet in develop

We don't need the match server here, since we are using `devNet` again.

alastairong:
Oh you're using devnet-resolver?

zo-el:
This branch has just been updated to be used as a QA testing branch

PR #1284, opened 2022-06-29
membrane-proof-service: can register multiple agents only for hosts roles

zo-el:
This is to add a check such that and registration code for any role other than `host` can only use the registration code for one agent.

Note: We have a requirement that the hosts should be able to use their one registration-code to register multiple agents. This is why the mem-proof did not have this check-up till now.


robbiecarlton (approved):
Looks good.

Are we no longer using this repo:

https://github.com/Holo-Host/membrane-proof-service

That repo has some different logic around "permissions" that seems to be trying to solve the same problem

JettTech (approved):
LGTM

overlays/holo-nixpkgs/membrane-proof-service/src/helper/register-user.js
```diff
@@ -25,6 +25,24 @@ async function registerUser(res, payload) {
 
     console.log("CodeList: ", codeList)
     console.log("codeList.agentPubKeys: ", codeList.agentPubKeys)
+    
+    // Role = Host: lets you register multiple agents under the same registration code
+    // For the rest of the roles we need to let them register only one agent
+    if ((role !== "host") && (codeList.agentPubKeys.length > 0)) {
+        if (codeList.agentPubKeys[0].pubKey == agent_pub_key) {
```
JettTech, line 32:
```suggestion
        if (codeList.agentPubKeys[0].pubKey === agent_pub_key) {
```

PR #1308, opened 2022-08-03
hpos-holochain-api & admin: new endpoints

JettTech:
### Updates:
 - add `/host_earnings` endpoint
 - add `/zome_call` endpoint
 - renames `/dashboard` endpoint to `/usage` +  adds functionality & return values
 - adds test setup for `/host_earnings` (pre-seeds hf and servicelogger DNAs)
 - adds tests for `/zome_call` and `/host_earnings`
  - updates `/usage` test
  - restructure build setup to use es6


overlays/holo-nixpkgs/hpos-holochain-api/src/index.js
```diff
@@ -18,18 +22,77 @@ const {
...
+const { parsePreferences, getHostedHappDetails, getEarningsForDays, isUsageTimeInterval } = require('./utils')
 const { getAppIds, getReadOnlyPubKey } = require('./const')
-const { AdminWebsocket, AppWebsocket } = require('@holochain/client')
 const { register_happ } = require('./handler/register_happ')
 const { inspect } = require('util')
 
+const getHostEarnings = async () => {
+	const appWs = await AppWebsocket.connect(`ws://localhost:${HAPP_PORT}`)
+	const APP_ID = await getAppIds()
+	const completed_txs = await callZome(
```
zeeshan595, line 33: This can be improved to do all 3 of these calls in parallel. 
Like this:
```ts
const getTransactions = async () => {
  const completed_txs = callZome(
    appWs,
    APP_ID.CORE,
    'holofuel',
    'transactor',
    'get_completed_transactions',
    null
  );

  const pending_txs = callZome(
    appWs,
    APP_ID.CORE,
    'holofuel',
    'transactor',
    'get_pending_transactions',
    null
  );

  const actionable_txs = callZome(
    appWs,
    APP_ID.CORE,
    'holofuel',
    'transactor',
    'get_actionable_transactions',
    null
  );

  return await Promise.all([
    completed_txs,
    pending_txs,
    actionable_txs
  ]);
}
```
That will return an array with 3 elements containg
`[completed_txs, pending_txs, actionable_txs]`
but importantly it will all happen togather instead of 1 at a time

overlays/holo-nixpkgs/hpos-holochain-api/tests/helpers/complete_hosting_invoice.js
```diff
@@ -0,0 +1,107 @@
...
+    const invoice = await callZome(
+		appWs,
+        APP_ID.CORE,
+        'holofuel',
+        'transactor',
+        'create_invoice',
+        invoiceArgs
+	)
+
+	await wait(12000)
```
zeeshan595, line 51: what is the purpose of this wait for 12 seconds, I am assuming it is in miliseconds?

If we are awaiting to receive back a message I would prefer an event based approch where once we get the message it triggers us to go to the next step.

zeeshan595 (changes requested):

lair/lair-keystore-config.yaml
```diff
@@ -0,0 +1,41 @@
+---
+
+# The connection url for communications between server / client.
+# - `unix:///path/to/unix/socket?k=Yada`
+# - `named_pipe:\\.\pipe\my_pipe_name?k=Yada`
+# - (not yet supported) `tcp://127.0.0.1:12345?k=Yada`
+connectionUrl: "unix:///home/zo-el/Documents/gitrepos/holochain/holo/hpos-services/holo-nixpkgs/lair/socket?k=Q46umK-8KzhRRabEPCtgnG9lDBAqQMDaQQ6LhoV9NyA"
```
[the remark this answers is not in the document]
alastairong: Haha zo-el is a universal user? :)

overlays/holo-nixpkgs/hpos-holochain-api/tests/helpers/complete_hosting_invoice.js
JettTech: Thanks for the feedback.  Yes, it is ms. :sweet_sweat:  Also, we're just doing a timeout at the moment, since we don't have the tooling exposed to make a state dump and ensure that the update has been propagated.  As this is a distributed system, there is uncertainty about when the update has been received by the another user - the only way to know that a user has the update without here being a built-in method in the DNA, is by checking their chain.  Oringally, I was thinking about just going with the lazy approach of timeouts - as this is what has been used throughout some of our DNA testing whenever state dump wasn't exposed, but I can look into how much of a lift it is to add the state dump helper here as well.

overlays/holo-nixpkgs/hpos-holochain-api/src/index.js
JettTech: Good point. Thanks! :D

JettTech: The structure of the return values is different between `completed_tx` and the other two -- `completed_tx` returns an array of objects, while the other two return an object whose values are arrays of objects.  Combining them all would make it a bit messier to clean and filter.  That said, we could def combine the latter two - `pending_txs` and `actionable_txs`  - and thus still optimize this a bit by calling these two in parallel.  Thoughts?

zeeshan595: You don't need to combine them. 
`Promise.all` will always return in order So if the first promise is 
```ts
callZome(
    appWs,
    APP_ID.CORE,
    'holofuel',
    'transactor',
    'get_completed_transactions',
    null
  );
  ```
  then the first element in the array would be `completed_txs`.

  If you prefer you can write it like this.
```ts
const completed_txs_p = callZome(
  appWs,
  APP_ID.CORE,
  'holofuel',
  'transactor',
  'get_completed_transactions',
  null
);

const pending_txs_p = callZome(
  appWs,
  APP_ID.CORE,
  'holofuel',
  'transactor',
  'get_pending_transactions',
  null
);

const actionable_txs_p = callZome(
  appWs,
  APP_ID.CORE,
  'holofuel',
  'transactor',
  'get_actionable_transactions',
  null
);
const [
  completed_txs, 
  pending_txs, 
  actionable_txs
] = await Promise.all([
    completed_txs_p,
    pending_txs_p,
    actionable_txs_p
]);
```

lair/lair-keystore-config.yaml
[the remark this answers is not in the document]
JettTech: Thanks for pointing this out - this is testing residue.  I'll remove it.

overlays/holo-nixpkgs/hpos-holochain-api/src/index.js
JettTech: Got it. For some reason, I was thinking that `Promise.all` returned it as: `[...foo, ..bar]`, instead of `[foo, bar]`.  This will work then and is a good optimization.  Updating now. :)

overlays/holo-nixpkgs/hpos-holochain-api/tests/helpers/complete_hosting_invoice.js
zeeshan595: Thought I would just mention it here that I am not a huge fan of this approach since it can lead to race conditions and generally unpredictable behaviour if something is quicker than expected or takes longer than we thought it would. Then the system can end up in a broken state. 

I would much prefer event driven approach where a function is triggered when a event happens rather than we wait for the event to happen and hope that it happened within the time we specified.

I am going to approve the PR since this is already in the code base and maybe requires a bigger discussion going forward?

cc @alastairong

zeeshan595 (approved):

overlays/holo-nixpkgs/hpos-holochain-api/tests/helpers/complete_hosting_invoice.js
alastairong: > since it can lead to race conditions and generally unpredictable behaviour

Distributed systems in a nutshell? ;)

zeeshan595 (approved):

PR #1310, opened 2022-08-09
Change services.configure-holochain option from holo-network to force-random-agent-key

peeech:
Currently `holo-network` nixOs variable is passed to `configure-holochain.service` via `services.configure-holochain.holo-network` [setting](https://github.com/Holo-Host/holo-nixpkgs/blob/develop/modules/services/configure-holochain.nix#L17). `Configure-holochain` consumes it as an env var in [this line of code](https://github.com/Holo-Host/hpos-configure-holochain/blob/develop/src/websocket.rs#L58) to decide what agent key is to be used.

The flaw of this architecture is that `holo-network` logic is moved to the code of a service. What `configure-holochain` binary needs to know is in fact whether it should use agent key from the config file or create a new random one. As a consequence variious hacks need to be used (i.e. in [memproof-server profile](https://github.com/Holo-Host/holo-nixpkgs/blob/develop/profiles/logical/holo/membrane-proof-server/default.nix#L130)) that break the logic of `holo-network`.

## Proposed change

If we change `services.configure-holochain.holo-network` to `services.configure-holochain.force-random-agent-key` then we can explicitely tell `configure-holochain.service` (i.e. in memproof-server profile) what to do with agent keys.


peeech:
I am not sure what logic is governing value of `force-random-agent-key` in those 3 profiles:
- https://github.com/Holo-Host/holo-nixpkgs/blob/config-hc-networks/profiles/logical/holo/holo-business-services/default.nix#L154
- https://github.com/Holo-Host/holo-nixpkgs/blob/config-hc-networks/profiles/logical/holo/membrane-proof-server/default.nix#L130
- https://github.com/Holo-Host/holo-nixpkgs/blob/config-hc-networks/profiles/logical/holo/match-server/default.nix#L35

peeech:
@zo-el Would you be able to describe logic for those three?

zo-el:
Based on the recent requirements, we need the holo servers to have permanent agent keys. So we need to restructure this code anyway. 

We don't necessarily need a `force-random-agent-key`(we can still have that var) as much as we need a `read-from-server-config`. 

America already has this task on the board and work is started on it. So I would recommend just leaving this in its current state and we have a clean PR to develop to clean up this problem.

peeech:
Ok, we can close this PR, but my suggestion is to take from this solution a part, that all the logic regarding key management behaviour is in profiles which would pass an outcome of a logic in form of a boolean to `configure-holochain` @zo-el

PR #1319, opened 2022-08-22
o/hn/holochain: cherry-pick b7c2fd10497d9cd883d09b2fae22629aee110137

zo-el:


alastairong:
Adding context to this PR.

This PR disables rate-limiting in Holochain. Rate-limiting is a feature intended to prevent infinite loops by setting a maximum number of wasm ops per zome call. There is a known bug with rate-limiting in this version of Holochain.

This is a temporary workaround that will be removed when we bump to a new version of holochain with the bug removed.

PR #1328, opened 2022-09-05
Holofuel UI served form HPOS in self hosted context

peeech:
- [x] Allows to server holofuel UI from holoport in a self hosted context
- [x] Adds integration with Host Console
- [x] Switches `hp-admin-crypto` to tokne-based authentication (see [this doc](https://hackmd.io/Evi5CVFqTI22UD8_OrDgYA?view))


peeech:
Also supersedes #1274

